PR #335, opened 2014-02-03
The install instructions state 2.0

BonnieDoug:
The install instructions state 2.0 is required of Admin and Extends bundles however they are both now on 2.1 and so is SonataUser Bundle Is this correct?


OskarStark:
> I think I've miss understood how you use github. Could you please move this to issue (if it is an issue). Sorry.

PR #413, opened 2014-05-24
removing setCurrentUri call from getMenu-Method

leberknecht:
since [this commit](https://github.com/KnpLabs/KnpMenu/commit/e7ef14a9469545423276e8516196496664429b81), setCurrentUri is gone (and if i get it right, url-matching is now done outside of the menu object)


roman-1983:
This issue is related to https://github.com/sonata-project/sandbox/issues/484

Soullivaneuh:
Already fixed: https://github.com/sonata-project/SonataUserBundle/blob/ec09378a100dfe6285562e864c2286ef9104059e/Block/ProfileMenuBlockService.php#L87

PR #414, opened 2014-05-25
fixing build - step 1 - adjusting composer.json

leberknecht:
Since [87f2072](https://github.com/sonata-project/SonataUserBundle/commit/87f2072) we require PageableManagerInterface, but the interface was added in core-bundle [4f48318](https://github.com/sonata-project/SonataCoreBundle/commit/4f48318), which is beyond what we get with 
>"sonata-project/core-bundle": "~2.2"

(composer will checkout 2.2.5)
So to green the build, we need to checkout dev version :/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "symfony/form": "~2.3",
         "symfony/security": "~2.3",
         "symfony/console": "~2.3",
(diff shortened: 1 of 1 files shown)


OskarStark:
i will close this PR, feel free to reopen

PR #417, opened 2014-05-27
Fixes #412 and a possible niche issue

### LINKED ISSUE #412: On login page, register form gets embedded in base layout (resulting in two nested base layout)
While trying to figure why my login page would have two nested base layouts, I narrowed it down to `SonataUserBundle / Resources / views / Security / base_login.html.twig`, which calls:

```
{% render controller("SonataUserBundle:RegistrationFOSUser1:register") %}
```

But then the controller renders the form by calling:

```
        return $this->container->get('templating')->renderResponse('FOSUserBundle:Registration:register.html.'.$this->getEngine(), array(
            'form' => $form->createView(),
        ));
```

which uses `{% extends "FOSUserBundle::layout.html.twig" %}`...

... resulting in nested base layouts...

Am I missing something or is there a bug somewhere?

net02:


OskarStark:
looks like this is already fixed, but feel free to reopen this PR if i missed something

PR #429, opened 2014-06-19
Update installation.rst

todstoychev:
Sonata user bundle needs sonata-project/datagrid-bundle for installation if @dev-master is used


OskarStark:
already done in `composer.json` : https://github.com/sonata-project/SonataUserBundle/blob/master/composer.json#L30

PR #434, opened 2014-07-03
Prevents BC break with KnpMenuBundle v1.x

LeoBenoist:
Like it was done in SonataAdminBundle https://github.com/sonata-project/SonataAdminBundle/blob/master/Admin/Admin.php#L1434


OskarStark:
why did you delete this branch? @LeoBenoist

LeoBenoist:
It was more than on year old and suppose at this point if it hasn't been merged or commented yet it will not be. Is that of any interest ?

PR #436, opened 2014-07-10
Removing deprecated method

mikeSimonson:
The setCurrentUri method has been removed a long time ago.
https://github.com/KnpLabs/KnpMenu/commit/e7ef14a9469545423276e8516196496664429b81

Right now If your follow the install instruction it breaks when you try to go on the profile page because it use that method.

### DIFF
--- a/Block/ProfileMenuBlockService.php
+++ b/Block/ProfileMenuBlockService.php
@@ -87,6 +87,10 @@ class ProfileMenuBlockService extends MenuBlockService
             if (method_exists($menu, 'setCurrentUri')) {
                 $menu->setCurrentUri($settings['current_uri']);
             }
+
+	    if (method_exists($menu, 'setUri')) {
+                $menu->setUri($settings['current_uri']);
+            }
         }
 
         return $menu;


Schyzophrenic:
Thanks @mikeSimonson Any idea if this PR will be merged one day ?

Soullivaneuh:
That will be BC break, isn't it?

Could you please rebase your PR and add some test revealing this issue?

mikeSimonson:
I cannot really add tests because it just blow up if you try to use a method that does not exists anymore.

Soullivaneuh:
Well, if this method does not exists anymore, just making a test that will execute `ProfileMenuBlockService::getMenu` will throw the needed fail.

Seems to be not coverage actually.

mikeSimonson:
Anyway it seems to be corrected in master and I don't have the test setup I had to check that anymore.

Soullivaneuh:
@mikeSimonson Could you please rebase your PR?

mikeSimonson:
The problem is that my PR fixed the issue in the sense that it didn't blew up anymore but as I don't know what it was supposed to do exactly and it isn't tested so I have no way to know if I break a part of the functionality or not.

Soullivaneuh:
See with @rande. ;-)

roman-1983:
Would be great if this issue will be fixed. I ran into the same error as @mikeSimonson

Block/ProfileMenuBlockService.php
```diff
@@ -87,6 +87,10 @@ protected function getMenu(BlockContextInterface $blockContext)
             if (method_exists($menu, 'setCurrentUri')) {
                 $menu->setCurrentUri($settings['current_uri']);
             }
+
+	    if (method_exists($menu, 'setUri')) {
```
Soullivaneuh: Please use a `if/else` statement with a comment for which version you BC fix.

OskarStark:
@mikeSimonson are you willing to finish this PR?

mikeSimonson:
@OskarStark I have no way to be sure that I am not breaking any functionality.

OskarStark:
thanks for your feedback!

PR #440, opened 2014-07-16
BugFix of mongoDb support

Dropaq:
It's a fix of severe bug.

Support of mongoDB by Sonata-User-Bundle was completely broken, because Doctrine was unable to find `FOS\UserBundle\Resources\config\doctrine\User.mongodb.xml` while it iterates through inheritance.

So basically **FOSUserBundle** has `FOS\Model` which is extended twice by `FOS\Entity` and by `FOS\Document`.

At the same time **SonataUserBundle** implementing similar logic, but `Sonata\UserBundle\Model\User` is extending `FOS\Entity`.
It really doesn't make any sense. The only reason of existence of this Model is preventing from copy-pasting common fields between Entity and Document, but the price of it is too high.

Since support of PHP 5.3 is necessary and usage of traits is impossible, I just copied methods from Entity into Document.
Also I've fixed logic of `UserManagerInterface` and `UserManagerProxy` which caused similar problem with serializer. So basically missed declaration of service for mongoDB resulted in Fatal Error on project update.
By now all methods of newly updated `UserManagerInterface` for `Document\UserManager` is replaced with stubs throwing an Exception. I suppose serialization wouldn't work for mongo, but looks like everything else is up an running.

PR #441, opened 2014-07-16
 BugFix of mongoDb support for 2.2 brunch

Dropaq:
Because we need this fix in stable version for our project, it is the same PR as: https://github.com/sonata-project/SonataUserBundle/pull/440 but created for 2.2 brunch (since few parts in Model was changed since it was released)


Soullivaneuh:
PRs are only accepted on master for the moment.

PR #467, opened 2014-09-09
Update GroupAdmin.php

cassianotartari:
Without the text in the 'with' method the symfony was throwing:

ContextErrorException: Warning: Missing argument 1 for Sonata\AdminBundle\Mapper\BaseGroupedMapper::with(), called in /var/www/html/muva/vendor/sonata-project/user-bundle/Admin/Model/GroupAdmin.php on line 68 and defined in /var/www/html/muva/app/cache/dev/classes.php line 12165

### DIFF
--- a/Admin/Model/GroupAdmin.php
+++ b/Admin/Model/GroupAdmin.php
@@ -65,7 +65,7 @@ class GroupAdmin extends Admin
                 ->end()
             ->end()
             ->tab('Security')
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Seems to be already done: https://github.com/sonata-project/SonataUserBundle/blob/master/Admin/Model/GroupAdmin.php#L68

Issue #469, opened 2014-09-11
Problems with RestoreRolesTransformer


vstrizhenok:
line 67 must contain this line
 $hiddenRoles = array_diff($this->originalRoles, array_merge($availableRoles, array_keys($availableRoles)));

because in case when used hierarhical roles, the transformer works wrong. Admin user can't edit another users with some roles that he (admin) has throw hierarchy. 
For example we need to add/remove role ROLE_COMPANY_PERSONAL_MODERATOR to the user `Pete`. Our admin has role ROLE_COMPANY_ADMIN that contains ROLE_COMPANY_PERSONAL_MODERATOR. So, when we try to do this we will fail because variable $availableRole contains line 'ROLE_COMPANY_ADMIN: ROLE_COMPANY_PERSONAL_MODERATOR' (it is wrong, because array must contains role NAME!). But the variable $this->originalRoles contains line 'ROLE_COMPANY_PERSONAL_MODERATOR' (it is right). So, in the result of the array_diff execution we have wrong result, that role 'ROLE_COMPANY_PERSONAL_MODERATOR' is hidden. 

So I suggest to use array_key instead values (if this transformer used only for form type sonata_security_roles) or merge array_keys of the $avaliableRoles array with its values in another case.

pamuche:
fixed & merged in master: #522

PR #470, opened 2014-09-12
Fixing problem with RestoreRolesTransformer #469

vstrizhenok:
I've made changes by first way of my suggestion
"So I propose to use array_key instead values (if this transformer used only for form type sonata_security_roles) or merge array_keys of the $avaliableRoles array with its values in another case."


OskarStark:
This is fixed in the current master!

Thank you for your contribution (y)

PR #490, opened 2014-10-24
Replace url() with path() for logout url

samuel4x4:
It's not better to use the relative logout url generated with the path() twig function, instead of the absolute url using the url() function?

I noticed in a project of mine (the hosting server have some special settings regarding the public hostname) that the logout url is wrong generated using url(), with path() it is perfect.


Soullivaneuh:
Please rebase your PR.

PR #495, opened 2014-11-21
Cleaning up. Past tense of a verb isn't needed here.

micheal:


Soullivaneuh:
Wrong. This is the actual saying is good.

See:

* http://able2know.org/topic/63527-1
* http://en.bab.la/conjugation/english/split

Feel free to reopen it and add an argument if I'm wrong. :+1:

Soullivaneuh:
And BTW, we don't accept PR only on master for the moment.

Issue #512, opened 2015-02-04
Variable "admin" does not exist in SonataUserBundle:Admin/Core:user_block.html.twig at line 7


grizzlylab:
Hi, got this error just after the login form. 
To fix it just replace the line 7 by: 
 {% set _edit_profile_uri  = url('admin_sonata_user_user_edit', {'id':app.user.id}) %}

Bladrak:
Should be fixed by #514

PR #513, opened 2015-02-05
Variable admin does not exist in SonataUserBundle:Admin/Core:user_block.html.twig at line 7

lumilo8:
#512


Bladrak:
@lumilo8 this won't solve the problem if you have a custom namespace in your user class override (the route name will then be different)

lumilo8:
Understood, thank you @Bladrak, i'm going to close this

Issue #461, opened 2014-08-21
Unable to remove user roles, if role has hierarchy


LukasLt:
We were unable to remve assigned user roles, like `ROLE_BUYER: ROLE_USER`, using Sonata Admin bundle.

We nailed down to `Sonata\UserBundle\Form\Transformer\RestoreRolesTransformer::reverseTransform() (65)`

       list($availableRoles, ) = $this->rolesBuilder->getRoles();
       $hiddenRoles = array_diff($this->originalRoles, $availableRoles);

The problem is that `$this->originalRoles` user 'keys', while `$availableRoles` - 'values'. 

     var_dump($this->originalRoles);
     #     array(2) {
     #       [0]=>
     #         string(10) "ROLE_BUYER"
     #         <...>
     #     }

     var_dump($availableRoles);
     #     array(12) {
     #       [ROLE_BUYER]=>
     #         string(10) "ROLE_BUYER: ROLE_USER"
     #         <...>
     #    }

Fix was add array_keys as such

     $hiddenRoles = array_diff($this->originalRoles, array_keys($availableRoles));

otherwise, `ROLE_BUYER` was allways readded as hidden role again :/

pamuche:
+1 for merging the solution from @LukasLt (https://github.com/sonata-project/SonataUserBundle/issues/461#issue-40799219) into master.

pamuche:
Ok, found thats allready fixed & merged in master:
https://github.com/sonata-project/SonataUserBundle/pull/522/

Soullivaneuh:
Thanks for the tip @pamuche

Issue #499, opened 2014-12-04
Hidden roles are not kept after edit


edorfaus:
If a user can edit some of the roles of a group, but not all of them, then the ones the user does not have access to are hidden (not shown in the role list). If one of those hidden roles were originally selected for that group, after this user clicks update, that role will be removed from the group. Since it was never shown to the user, it should instead be kept selected.

This problem was introduced by PR #452 - which causes SecurityRolesType to set the transformer's list of original roles to the list of roles that were selected by the user instead of setting it to the list of roles that the group had originally.

Therefore, I think that change should be reverted, and the problem it worked around (issue #423) should be fixed in a different way - I would suggest using the array_keys solution mentioned in issues #497, #461 and #469, and implemented in pull requests #471 and #470 by @vstrizhenok (which also include a test).

mkurzeja:
@Bladrak Any change to fix this issue? Please let me know if preparing a single PR that reverts changes from #452 and includes #470 would help. I could submit it with some additional tests.

Bladrak:
@mkurzeja no change on my side, but feel free to submit the PR. @rande may check this out

PR #522, opened 2015-02-20
Fixing hidden roles support in SecurityRolesType

mkurzeja:
As mentioned in #499 there is a problem with keeping hidden roles.

Example:
User 'admin' with roles: ROLE_ADMIN, ROLE_USER is editing user 'superadmin' with ROLE_SUPER_ADMIN.

Currently Sonata would delete ROLE_SUPER_ADMIN from superadmin user. Expected behavior is to keep this role.

I reverted the change from #452 and added changes from #470. I didn't delete tests from both PR's and added one additional tests to cover the error from #499.

### DIFF
--- a/Form/Transformer/RestoreRolesTransformer.php
+++ b/Form/Transformer/RestoreRolesTransformer.php
@@ -64,7 +64,7 @@ class RestoreRolesTransformer implements DataTransformerInterface
 
         list($availableRoles, ) = $this->rolesBuilder->getRoles();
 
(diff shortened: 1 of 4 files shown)


mkurzeja:
@rande any chance to get this merged? Let me know if I should change anything

pamuche:
Also fixed #461

kaiwa:
@rande I'm missing this patch in the 2.2 branch. Don't get bugfixes merged into the current release?

PR #543, opened 2015-04-17
Moved method definition to interface

core23:


Soullivaneuh:
Which interface you mean? Lot of modified methods does not have parents.

And please use `{@inheritdoc}` instead of `@inheritdoc`

core23:
Some methods are only defined in the concrete user class, would be better to move them to the interface. 

What do you mean with "Lot of modified methods does not have parents"?

core23:
@Soullivaneuh ping

Soullivaneuh:
> What do you mean with "Lot of modified methods does not have parents"?

Nothing. My IDE just drank wine.

But, adding method to interface is BC break.

If this could not be handled, this should be merge only on next major release.

BTW, an upgrade note should be inserted.

core23:
Changelog should be fine now

PR #544, opened 2015-04-22
Makes user expiresAt nullable

GromNaN:
After adding the user expiration date to the admin, the field becomes mandatory.

This is an alternative to https://github.com/FriendsOfSymfony/FOSUserBundle/pull/1812

### DIFF
--- a/Model/User.php
+++ b/Model/User.php
@@ -180,6 +180,20 @@ abstract class User extends AbstractedUser implements UserInterface
     }
 
     /**
+     * Set expiration date
+     *
+     * @param \DateTime|null $date
+     *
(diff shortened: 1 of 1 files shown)


Model/User.php
```diff
@@ -180,6 +180,20 @@ public function getExpiresAt()
     }
 
     /**
+     * Set expiration date
+     *
+     * @param \DateTime|null $date
```
eko: Be careful, there is a wrong parameter name here, it should be: `$expiresAt`

GromNaN: Fixed. Thanks.

Issue #322, opened 2014-01-22
Support for FosUserBundle 2.X


rande:
This issue regroups information about support FosUserBundle 1.x and 2.x.

numerogeek:
+1

limfinity:
+1

Tetdoss:
+1

lumilo8:
We have the same problem... integrating https://github.com/PUGX/PUGXMultiUserBundle it requires fosuser 2.0 :(
:+1:

VolodymyrChornohal:
@rande hi! what version or release of SUB should I try in order to satisfy FOSUB2.0 composer requirements? I've checked all the releases and tagged versions but the requirement is always FOSUB1.3 :(

extdevelopment:
+1

OskarStark:
:+1:

greg0ire:
Please stop the "+1", "me too" comments.

rande:
@greg0ire a PR is more efficient ;)

greg0ire:
Yeah definitely :) Also, it would be great if github had a built-in +1 feature so that people could vote for issues without leaving the comment and notifying everyone in the thread. @everyone : Please **do not** +1 that comment if you agree.

pulse00:
I'm working on a PR for [`fos-user-bundle/2.x`](https://github.com/pulse00/SonataUserBundle/tree/feature/fos-2.x) support. 

I have a question regarding the idea behind naming the controllers `**FOSUser1Controller`. I've currently simply added a second set of controllers/routings using `**FOSUser2Controller` which work with the fos-user 2.x version. Do i understand it right that both 1.x and 2.x versions of the FUB should be supported?

What about the version constraint in `composer.json`, should it be set to something like `>=1.0,<=2.0` ?

phansys:
IMO, ```"friendsofsymfony/user-bundle": "~2.0"``` would be great for FOSUB 2.x support.

pulse00:
@rande any input on this would be great.

roman-1983:
+1

rande:
@pulse00 sorry, I didn't see your comment ;)

rande:
@pulse00 the FosUserBundle 2.0 is not yet stable, but people is using it. So yes I would like to support both version.

pulse00:
@rande alright, thanks for the info

numerogeek:
@rande @pulse00 hi guys, any news about this issue ? 
[...]

PR #555, opened 2015-05-21
Feature/fos 2.x

pulse00:
This PR adds support for the current 2.x branch of the FOSUserBundle. The main changes:

- Added `FOS*2Controller` controllers
- Added `*_2.xml` routing configuration
- Added documentation

The bundle will work with both version of the fos-user-bundle, the only things which needs to be changed are:

- Set the `friendsofsymfony/user-bundle` version accordingly in `composer.json`
- Use the `*2.xml` versions of the routing configuration

I've created 2 demo apps which use either the 1.x or the 2.x version of the fos-user-bundle:

- https://github.com/pulse00/sonata-user-fos2
- https://github.com/pulse00/sonata-user-fos1

Refs #322


Controller/RegistrationFOSUser2Controller.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Sonata\UserBundle\Controller;
+
+use FOS\UserBundle\Controller\RegistrationController;
+
+class RegistrationFOSUser2Controller extends RegistrationController
+{
+
+}
```
phansys: Missing single blank line at EOF.

Controller/ResettingFOSUser2Controller.php
```diff
@@ -0,0 +1,27 @@
...
+ * Class ResettingFOSUser1Controller
+ *
+ * @package Sonata\UserBundle\Controller
+ *
+ * @author Hugo Briand <briand@ekino.com>
+ */
+class ResettingFOSUser2Controller extends ResettingController
+{
+
+}
```
phansys: Missing single blank line at EOF.

Controller/SecurityFOSUser2Controller.php
```diff
@@ -0,0 +1,37 @@
...
+        if ($user instanceof UserInterface) {
+            $this->container->get('session')->getFlashBag()->set('sonata_user_error', 'sonata_user_already_authenticated');
+            $url = $this->container->get('router')->generate('sonata_user_profile_show');
+
+            return new RedirectResponse($url);
+        }
+
+        return parent::loginAction($request);
+    }
+}
```
phansys: Missing single blank line at EOF.

Resources/config/routing/sonata_change_password_2.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
+
+    <route id="fos_user_change_password" pattern="/change-password">
```
phansys: ```pattern``` is [deprecated](https://github.com/symfony/Routing/blob/v2.2.0/CHANGELOG.md#220) from Symfony 2.2. You should use ```path``` instead.

Resources/views/Security/base_login.html.twig
```diff
@@ -83,7 +83,7 @@ file that was distributed with this source code.
         </div>
         <div class="col-sm-6">
             {% block sonata_user_registration %}
-                {% render controller("SonataUserBundle:RegistrationFOSUser1:register") %}
+                {% render controller("SonataUserBundle:RegistrationFOSUser2:register") %}
```
rande: @pulse00 should we have 2 templates here or at least a variable to use version 1 or version 2 ?

Model/User.php
```diff
@@ -11,7 +11,7 @@
 
 namespace Sonata\UserBundle\Model;
 
-use FOS\UserBundle\Entity\User as AbstractedUser;
+use FOS\UserBundle\Model\User as AbstractedUser;
```
rande: Is this class available with FosUser 1 ?

pulse00: @rande yes: https://github.com/FriendsOfSymfony/FOSUserBundle/blob/1.3.x/Model/User.php

Controller/RegistrationFOSUser2Controller.php
pulse00: @phansys fixed https://github.com/pulse00/SonataUserBundle/commit/43ef7ce3d2de34077542cd320f9b3a9816f2b7cd#diff-d69f269adef698ce617762593ebb83dfR10

Controller/ResettingFOSUser2Controller.php
pulse00: @phansys fixed.

Controller/SecurityFOSUser2Controller.php
pulse00: @phansys fixed.

Resources/views/Security/base_login.html.twig
pulse00: @rande how would you determine fos 1/2 without a variable?

Controller/ProfileFOSUser2Controller.php
```diff
@@ -0,0 +1,155 @@
...
+                'breadcrumb_context' => 'user_profile',
+        ));
+    }
+
+    /**
+     * @return Response
+     *
+     * @throws AccessDeniedException
+     */
+    public function editProfileAction(Request $request)
```
phansys: Missing docblock for argument.

Resources/config/routing/sonata_change_password_2.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
+
+    <route id="fos_user_change_password" path="/change-password">
+        <default key="_controller">SonataUserBundle:ChangePasswordFOSUser2:changePassword</default>
+        <requirement key="_method">GET|POST</requirement>
```
phansys: ```_method``` requirement is deprecated. It should be:
```xml
<route id="fos_user_change_password" path="/change-password" methods="GET POST">
    <default key="_controller">SonataUserBundle:ChangePasswordFOSUser2:changePassword</default>
</route>
```

Resources/doc/reference/installation.rst
```diff
@@ -28,6 +28,17 @@ Enable the Bundle
...
     php composer.phar update
 
+.. note::
+    This bundle supports both the ``1.x`` and the ``2.x`` version of the ``FOS/UserBundle``.
+    You need to add the version to your composer.json manually
+
+
+.. code-block:: bash
+
+    php composer.phar require friendsofsymfony/user-bundle:2.0.*@dev  # for the 2.0 version of the FOSUserBundle
```
phansys: ```friendsofsymfony/user-bundle:~2.0@dev``` should be better while keep consistency with the constraint style in the next snippet.

composer.json
```diff
@@ -23,7 +23,7 @@
         "symfony/console": "~2.3",
         "sonata-project/core-bundle": "^2.3.2",
         "sonata-project/admin-bundle": "~2.3",
-        "friendsofsymfony/user-bundle": "~1.3",
+        "friendsofsymfony/user-bundle": "~1.3 || 2.0.*@dev",
```
phansys: It should be ```"friendsofsymfony/user-bundle": "~1.3|~2.0@dev"``` I think.

Controller/SecurityFOSUser2Controller.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * @param Request $request
+     * @return RedirectResponse|\Symfony\Component\HttpFoundation\Response
+     */
+    public function loginAction(Request $request)
+    {
+        $user = $this->getUser();
+
+        if ($user instanceof UserInterface) {
+            $this->container->get('session')->getFlashBag()->set('sonata_user_error', 'sonata_user_already_authenticated');
```
phansys: You could get the session from the request object.

phansys:
Thanks for your work on this feature @pulse00!
I've left some comments about CS, but I think you could ease those fixes with [php-cs-fixer](http://cs.sensiolabs.org/).

composer.json
Soullivaneuh: Just `~1.3|~2.0`. See this: https://github.com/sonata-project/SonataUserBundle/blob/master/composer.json#L52-L53

phansys: :+1:

Controller/ProfileFOSUser2Controller.php
pulse00: @phansys fixed.

pulse00:
> I've left some comments about CS, but I think you could ease those fixes with php-cs-fixer.

i've let it run on the whole codebase, but it basically modified every single file :)

Soullivaneuh:
@pulse00 

> i've let it run on the whole codebase, but it basically modified every single file :)

Not a problem. This will be applied later anyway, like this: https://github.com/sonata-project/SonataAdminBundle/pull/3038

cc @phansys

AntonioCS:
Any ETA on this? I want to use sonataUserBundle with FOSUserBundle 2.0 so that I can use PUGXMultiUserBundle!!
In the current state of things, If I want to use all these bundles I will have to downgrade to Symfony 2.1 (or something like that)

Soullivaneuh:
@AntonioCS Again, yelling with a lot of exclamation marks will not help the motivation. Please be polite.

> In the current state of things, If I want to use all these bundles I will have to downgrade to Symfony 2.1 (or something like that)

It's not fos-user cause neither sonata-admin because both support Symfony 2.*.

Your cited bundle doesn't: https://packagist.org/packages/pugx/multi-user-bundle

kallosz:
@AntonioCS you can always fork code and make it compatible with fosuserbundle@2.0 ;-)

Resources/views/Security/base_login.html.twig
phansys: I think you can trust in some service that is only defined in `2.x`, like [```fos_user.listener.authentication```](https://github.com/FriendsOfSymfony/FOSUserBundle/commit/95d57add6cdd7c535a3471bc750d3ea15e347ecf#diff-a431e3d5d9e6fd8d0548da7680427fb4R8) by instance.

pulse00:
@Soullivaneuh @rande i've had some time to look into finishing this PR, but i fear supporting both versions of the user-bundle at the same time will require ugly hacks, due to the following reasons:

1. The `User` class in the fos-user-bundle changed it's namespace from `FOS\UserBundle\Entity\User` to `FOS\UserBundle\Model\User`

2. Some `form.handler` services have been removed in fos 2.x, which is why i have simply commented them out in the current state of this PR: https://github.com/sonata-project/SonataUserBundle/pull/555/files#diff-850942b3ba24ab03a40aaa81b6152852R176 

I think it will probably be best to wait for the release of 2.x and then support multiple versions of fos-userbundle in different branches.

Soullivaneuh:
> and then support multiple versions of fos-userbundle in different branches.

Don't really like this approach.

I would prefer: Support both 1.x and 2.x on next minor release of sonata-user (2.x) with deprecation system and remove 1.x support for next major release.

I think you could do some trick with `class_exists` and `method_exists` to check FOS features and create deprecated method with `trigger_error`.

If this is **impossible**, then we should wait this PR for next major release.

phansys:
@pulse00, I'm very interested in have this PR merged ASAP. Can I help you with something?
Thank you in advance.

pulse00:
@phansys  it's basically the 2 issues mentioned [above](https://github.com/sonata-project/SonataUserBundle/pull/555#issuecomment-113883633). If you want i can add you as a collaborator to the fork and you can work on this. i don't have too much time this month :(

phansys:
Sure @pulse00, I'll try to take a look soon.

phansys:
@pulse00, what do you think about an adapter for your point 1? Is it something viable?

phansys:
After reading the [upgrade](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Upgrade.md#user-and-group-class) file from 1.x to 2.x in FOSUB, I think there is no issue at all with the change of the inheritance from [FOS\UserBundle\Entity\User](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/1.3.x/Entity/User.php#L16) to [FOS\UserBundle\Model\User](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Model/User.php#L23) since the last was already in the inheritance chain.
At the worst case, if we consider this a "no-go" BC break, I think the removed classes from FOSUB can be recreated here.
Any thoughts on this?

phansys:
@pulse00, I've forked your repo, rebased the changes against master and added [an example](https://github.com/sonata-project/SonataUserBundle/compare/master...phansys:fosub2?expand=1#diff-52d0a1a5f6ea117229b7bf98406e9ba3R30) on how I think the previous handlers could be re-created.
Please, tell me what you think about this approach.

pulse00:
@phansys maybe you can take over the migration and create a separate PR? unfortunately i don't have much time atm, and the upgrade doesn't have a very high priority to us. 

/cc @Soullivaneuh

phansys:
Thank you for your feedback @pulse00. I'll try to work on this.

SchizoDuckie:
I just tested @phansys's implementation locally and it works like a charm. I would vote to merge it if there's no more problems detected.

I needed this because i want to capture some of the new FOSUser 2.0 hook events and forward them to Kibana.

For those interested: This is what I changed in my composer.json to get this UserBundle 2.3.x based on FosUserBundle 2.0.x to interact with the Sonata Adminbundle 2.3.4

composer.json
```
 "require": {
...
        "sonata-project/user-bundle": "dev-fosub2 as 2.3",
        "sonata-project/formatter-bundle": "~2.3",
        "knplabs/knp-markdown-bundle": "1.2.0",
        "egeloen/ckeditor-bundle": "~2.0"
...
},
"repositories": [{
        "type": "vcs",
        "url": "https://github.com/phansys/SonataUserBundle.git"
    }],
```

config.yml:
add sonata_formatters config

AppKernel.php:
```
...
            new Sonata\UserBundle\SonataUserBundle('FOSUserBundle'),
            new Sonata\FormatterBundle\SonataFormatterBundle(),
            new Knp\Bundle\MarkdownBundle\KnpMarkdownBundle(),
            new Ivory\CKEditorBundle\IvoryCKEditorBundle(),
...
```

routing.yml:
```
#sonata user
sonata_user:
    resource: '@SonataUserBundle/Resources/config/routing/admin_security.xml'
    prefix: /admin

sonata_user_security:
    resource: "@SonataUserBundle/Resources/config/routing/sonata_security_2.xml"

sonata_user_resetting:
    resource: "@SonataUserBundle/Resources/config/routing/sonata_resetting_2.xml"
    prefix: /resetting

sonata_user_profile:
    resource: "@SonataUserBundle/Resources/config/routing/sonata_profile_2.xml"
    prefix: /profile

sonata_user_register:
    resource: "@SonataUserBundle/Resources/config/routing/sonata_registration_2.xml"
    prefix: /register

sonata_user_change_password:
    resource: "@SonataUserBundle/Resources/config/routing/sonata_change_password_2.xml"
    prefix: /profile
```

toinouu:
+1

pulse00:
In the current state, this PR will not work supporting *both* FOS 1.x *and* 2.x. So to avoid any further confusion, i'll close this PR. If anyone would like to work on FOS 2.x support, he/she is happy to fork the branch, i'll leave it there.

rande:
@pulse00 can you elaborate the issue you have ?

OskarStark:
is it really necessary to support both version?

2015-08-25 21:18 GMT+02:00 Thomas <notifications@github.com>:

rande:
@OskarStark tricky question ;) If you look at the trend, there is 3K daily install for the dev-master https://packagist.org/packages/friendsofsymfony/user-bundle/stats#dev-master and 200 install for the stable version. Depends on what you want to support: stable vs dev (breakable release)

OskarStark:
i would recommend to use the 2.0 versions for further development and then
later use the stable version when it exists

but... i use the master of sonata too, instead of the stable 2.3... :)


2015-08-25 21:26 GMT+02:00 Thomas <notifications@github.com>:

pulse00:
> @pulse00 can you elaborate the issue you have ?

Non-technical issue is that i'm not working on PHP projects atm. The technical issues which block this PR are described here https://github.com/sonata-project/SonataUserBundle/pull/555#issuecomment-113883633

toinouu:
What is the interest of keeping backwards compatibility ? 
someone who wants to use fos 1.3 should just use a sonata-user =< 2.2.4

Where is the problem ?

OskarStark:
looks like the problem for now is, that there is no 2.0 STABLE of FOSUser, right?

toinouu:
Yes but if fos user 2.0 il stable in 2018, we must wait 2018 ?

OskarStark:
i agree with you @toinouu 

https://github.com/sonata-project/SonataUserBundle/pull/555#issuecomment-134708829

JHGitty:
There is currently not support for FOSUserBundle in version 2? That because I need to rewrite everything now to FOSUserBundle 1.x .. that is not good for our business. Hope for a fork which supports version 2 or official support.

OskarStark:
@JHGitty now there are 3 PreReleases of 2.0 so i think we should use the work of @pulse00 and make the current sonata dev working with the 2.0.0 alpha 3. later on we can switch to the 2.0 final version

@pulse00 could you help us with this an get rid of the FOS 1.0-Stuff in your PR?

rande:
For the record, there is a branch with @pulse00's work: https://github.com/sonata-project/SonataUserBundle/tree/add_support_for_fos_user2

SchizoDuckie:
I'm sorry, but all of this is a freaking mess of hacks upon patches upon forks upon half finished pull requests without docs.

I've left Sonata (which is was quite enthousiastic about at first) for what it is and moved on to other projects that have their dependency hell in order.

There is just no way that i can convince my boss or my colleagues to use any of the Sonata packages in the state they have been in for the last year.

rande:
@SchizoDuckie thanks got the feedback.

AntonioCS:
@SchizoDuckie What did you move to? If there is a project out there that provides more or less the same functionality of sonata without the problems I will move too :dancers:

SchizoDuckie:
@AntonioCS I've moved to Laravel5 with [this awesome boilerplate](https://github.com/rappasoft/laravel-5-boilerplate) and never looked back.

AntonioCS:
@SchizoDuckie I was hoping for something within symfony. Can't switch frameworks now :( Thanks anyway

PR #562, opened 2015-06-07
Improve true/false/null default values on configuration

Soullivaneuh:


Soullivaneuh:
https://travis-ci.org/sonata-project/SonataUserBundle/jobs/65758525

I don't think this is related with my PR.

PR #572, opened 2015-06-11
Fix lowest deps requirements

Soullivaneuh:


Soullivaneuh:
@rande I do not have this error on my local: https://travis-ci.org/sonata-project/SonataUserBundle/jobs/66349870#L406

Any idea?

rande:
@Soullivaneuh nop :/

Soullivaneuh:
@rande Well, the problem seems to be php `5.3.3` but dunno why and probably will never know. :-)

Also added `phpunit/phpunit` dev requirements to use the last version. Should be the same on other packages. :+1:

PR #589, opened 2015-06-25
Renamed two german strings

OskarStark:
and removed a blank line

### DIFF
--- a/Resources/translations/SonataUserBundle.de.xliff
+++ b/Resources/translations/SonataUserBundle.de.xliff
@@ -4,7 +4,7 @@
         <body>
             <trans-unit id="title_user_registration">
                 <source>title_user_registration</source>
-                <target>Benutzerregistrierung</target>
+                <target>Registrierung</target>
             </trans-unit>
             <trans-unit id="title_user_account">
                 <source>title_user_account</source>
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
I'm not speaking german. :-)

But can you please explain why your translation would be better?

Thanks.

OskarStark:
of course :)

ist some kind of double...

i'll try to explain it with english by side:

```Benutzerregistrierung``` in english is ```User registration```, but this is shorter:
```Registrierung``` - ```Registration```

so i stripped the ```Benutzer``` (```User```) of the translations, to make it shorter.

Soullivaneuh:
I see.

I'm ok for `title_user_registration` but I would like to keep `title_user_account` as is because it's the same way as the english translation: https://github.com/sonata-project/SonataUserBundle/blob/master/Resources/translations/SonataUserBundle.en.xliff#L11

A I right? If yes, could you please update it?

OskarStark:
@Soullivaneuh can you please merge, i reverted the change

Soullivaneuh:
Could you please squash your commits?

OskarStark:
@Soullivaneuh done

Soullivaneuh:
Thanks!

OskarStark:
np

thank you!

Issue #554, opened 2015-05-15
Sonata\UserBundle\Document\BaseUser is missing getGenderList method


emillosanti:
see Sonata\UserBundle\Entity\BaseUser 's implementation 

```
public static function getGenderList()
{
    return array(
        UserInterface::GENDER_UNKNOWN => 'gender_unknown',
        UserInterface::GENDER_FEMALE  => 'gender_female',
        UserInterface::GENDER_MALE    => 'gender_male',
    );
}
```

or just move to Sonata\UserBundle\Model\User

OskarStark:
can you please provide a PR then?

thanks

Soullivaneuh:
@emillosanti from which `UserInterface` these constant come from?

jordisala1991:
this should be solved on 3.x

PR #592, opened 2015-06-26
refactor User entity so that setters return $this

### LINKED ISSUE #541: User entity setters do not return $this
All the entity setters return void rather than $this.
While I understand that this is somewhat personal preference, I think they should return $this for the sake of consistency since the class extends FOSUserBundle's wherein all setters do return $this. Which means that when operating on a single $user instance some setters will return $this while others will not.

I can provide a PR if you're in agreement.

mcwebb:
fixes #541

### DIFF
--- a/Model/User.php
+++ b/Model/User.php
@@ -132,10 +132,14 @@ abstract class User extends AbstractedUser implements UserInterface
      * Sets the creation date.
      *
      * @param \DateTime|null $createdAt
+     *
+     * @return User
      */
     public function setCreatedAt(\DateTime $createdAt = null)
     {
         $this->createdAt = $createdAt;
+
+        return $this;
     }
 
     /**
@@ -152,10 +156,14 @@ abstract class User extends AbstractedUser implements UserInterface
      * Sets the last update date.
      *
      * @param \DateTime|null $updatedAt
+     *
+     * @return User
      */
     public function setUpdatedAt(\DateTime $updatedAt = null)
     {
         $this->updatedAt = $updatedAt;
+
+        return $this;
     }
 
     /**
@@ -192,10 +200,14 @@ abstract class User extends AbstractedUser implements UserInterface
      * Sets the credentials expiration date.
      *
      * @param \DateTime|null $date
+     *
+     * @return User
      */
     public function setCredentialsExpireAt(\DateTime $date = null)
     {
         $this->credentialsExpireAt = $date;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


OskarStark:
Looks good to me, but is this a BC?

Soullivaneuh:
@mcwebb Please replace `@return $this` by `@return User` and add a line before `return $this;`

Soullivaneuh:
> Looks good to me, but is this a BC?

@OskarStark I'm wondering the same question. I would think no because before that, nothing was returned.

But user dealing with non-result return won't have problem with this AFAIK.

Soullivaneuh:
@mcwebb Please also fix CS with `make cs` command.

See Travis job: https://travis-ci.org/sonata-project/SonataUserBundle/jobs/68440310#L396

OskarStark:
@Soullivaneuh yes that was also my thought, more should be not a problem.

@mcwebb if you are actively working on this, maybe you can have a look at this issue too:  #554

mcwebb:
I can't think of a context where this would be a BC; should be safe.
I'll sort out the issues over the weekend.

OskarStark:
OK, so when the CS is  fixed, it could be merged

Model/User.php
```diff
@@ -132,10 +132,12 @@
      * Sets the creation date.
      *
      * @param \DateTime|null $createdAt
+     * @return $this
      */
     public function setCreatedAt(\DateTime $createdAt = null)
     {
         $this->createdAt = $createdAt;
+        return $this;
```
phansys: [Please, add a blank line before ```return``` statements, unless the return is alone inside a statement-group](
https://symfony.com/doc/current/contributing/code/standards.html#structure).

Model/User.php
```diff
@@ -132,10 +132,12 @@
      * Sets the creation date.
      *
      * @param \DateTime|null $createdAt
+     * @return $this
```
phansys: >Annotations in phpdocs should be grouped together so that annotations of the same type immediately follow each other, and annotations of a different type are separated by a single blank line.
[[ref]](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/7ef7ea94ccb0d528daa2f428226958e00926b928/README.rst#usage)

Model/User.php
Soullivaneuh: Already said here: https://github.com/sonata-project/SonataUserBundle/pull/592#issuecomment-115573377

Soullivaneuh: BTW, just run `make cs` should correct it.

Soullivaneuh:
@phansys No need review here, already done. :+1:

phansys:
@Soullivaneuh, where is the appropriate place to do reviews now?

Soullivaneuh:
@phansys By default on the code as you did. But this issue is repeated on all the file so I just made a comment.

Comment considered by @mcwebb BTW: https://github.com/sonata-project/SonataUserBundle/pull/592#issuecomment-115630307

phansys:
Sure, I usually do comments in just one occurrence of the issue type, so I trust the author of the PR knows how to identify the rest of the issues.

Soullivaneuh:
ping @mcwebb

mcwebb:
I've fixed the cs and caught a nasty mistake I made in the original commit.
Should be all good now. Apologies for the delay.

phansys:
:+1: 
Indeed, I think this it's a kinda of bugfix since it allows to use the broken fluent [UserInterface](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/145b7167237ea70a1fb66956e8d96642c288e019/Model/UserInterface.php#L30) provided by FOSUB.

Soullivaneuh:
@mcwebb Thanks. Now please squash your commits.

mcwebb:
sorted

Soullivaneuh:
Thank you @mcwebb

PR #594, opened 2015-06-28
fixed crazy line breaks

OskarStark:


Soullivaneuh:
I'm globally agree with you for this proposition, but I prefer let @rande decide about it.

rande:
Thanks

PR #613, opened 2015-09-09
Security token can be null in login form on 404

wadjeroudi:
If you show the sonata login form in a 404 page, the token would be null and an error would show up.
Using the fos user action FOSUserBundle:Security:login doesn't throw an error.

It adds a check on the token.

### DIFF
--- a/Controller/SecurityFOSUser1Controller.php
+++ b/Controller/SecurityFOSUser1Controller.php
@@ -25,9 +25,9 @@ class SecurityFOSUser1Controller extends SecurityController
 {
     public function loginAction()
     {
-        $user = $this->container->get('security.context')->getToken()->getUser();
+        $token = $this->container->get('security.context')->getToken();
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Could you please add test that reveal the issue?

wadjeroudi:
You want me to do an unit test on a controller action ?
Or maybe a functional test ?

Soullivaneuh:
A unit test.

OskarStark:
any news @wadjeroudi

wadjeroudi:
@OskarStark @Soullivaneuh 
Sorry but I don't know how to do a unit test on a controller that depends on the whole container. If I instantiate the SecurityFOSUser1Controller and call the loginAction it fails on the first line.
For me we need a functional test for this case.

Soullivaneuh:

Regards.

PR #622, opened 2015-09-25
Better route name for sonata admin security

sagikazarmark:
I didn't find any reference to this route name, so I don't think it is actually a BC break.


OskarStark:
ah sorry, didn't catch that this was a doc change... my fault :)

sagikazarmark:
No problem. :+1:

sagikazarmark:
I still don't get why a simple doc change makes the tests failing. I guess there is something with the tests.

OskarStark:
Thank you @sagikazarmark

PR #626, opened 2015-09-28
Redirect to admin dashboard if already logged in

### LINKED ISSUE #619: Admin login bad redirection for already logged in user
When a user is already logged in and goes to `/admin/login` it should redirect to the dashboard. Instead it redirects to the profile page which is not just illogical (since you want to go to the dashboard), but also bad: in case frontend login is not allowed, it ends up with an error saying route `sonata_user_profile_show` cannot be found (in case frontend routes are not set).

sagikazarmark:
Fix #619

### DIFF
--- a/Controller/AdminSecurityController.php
+++ b/Controller/AdminSecurityController.php
@@ -27,7 +27,7 @@ class AdminSecurityController extends ContainerAware
 
         if ($user instanceof UserInterface) {
             $this->container->get('session')->getFlashBag()->set('sonata_user_error', 'sonata_user_already_authenticated');
-            $url = $this->container->get('router')->generate('sonata_user_profile_show');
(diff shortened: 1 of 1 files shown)


OskarStark:
Thank you @sagikazarmark

OskarStark:
yes it is @sagikazarmark but in the current master, not in the ``2.2``

@Soullivaneuh what do you think, can we cherry pick this commit and backport it to 2.2 and the create a release?

aderuwe:
@sagikazarmark @OskarStark thanks for passing this on, hope it's possible!

OskarStark:
@aderuwe could you please open a new issue, so that this need keep to stay in focus?

that would be awesome

aderuwe:
@OskarStark done!

PR #628, opened 2015-09-28
Fix deprecated settings method on block service

### LINKED ISSUE #576: Fix deprecated settings method on block service
To be done when https://github.com/sonata-project/SonataBlockBundle/pull/204 will be merged.

sagikazarmark:
Fix #576 

ping @Soullivaneuh


OskarStark:
hmm, you renamed the method, but i think this solution is correct:

https://github.com/sonata-project/SonataBlockBundle/pull/204/files

we should implement both methods, right @Soullivaneuh ?

Soullivaneuh:
No because the good method will be correctly called even if the block is extended with the old method usage.

See https://github.com/sonata-project/SonataAdminBundle/pull/3212 for example.

Soullivaneuh:
Thank you @sagikazarmark.

PR #629, opened 2015-09-29
Resolves path BC break

sagikazarmark:
Fixes sonata-project/SonataAdminBundle#3307

### DIFF
--- a/Controller/AdminSecurityController.php
+++ b/Controller/AdminSecurityController.php
@@ -64,12 +64,22 @@ class AdminSecurityController extends ContainerAware
             return new RedirectResponse($refererUri && $refererUri != $request->getUri() ? $refererUri : $this->container->get('router')->generate('sonata_admin_dashboard'));
         }
 
+        // TODO: Deprecated in 2.3, to be removed in 3.0
+        $resetRoute = 'sonata_user_admin_resetting_request';
+
+        // TODO: Deprecated in 2.3, to be removed in 3.0
+        if (null === $this->container->get('router')->getRouteCollection()->get('sonata_user_admin_resetting_request')) {
+            @trigger_error('Using the route fos_user_resetting_request for admin password resetting is deprecated since version 2.3 and will be removed in 3.0. Use sonata_user_admin_resetting_request instead.', E_USER_DEPRECATED);
+            $resetRoute = 'fos_user_resetting_request';
+        }
+
         return $this->container->get('templating')->renderResponse('SonataUserBundle:Admin:Security/login.html.'.$this->container->getParameter('fos_user.template.engine'), array(
                 'last_username' => $lastUsername,
                 'error'         => $error,
                 'csrf_token'    => $csrfToken,
                 'base_template' => $this->container->get('sonata.admin.pool')->getTemplate('layout'),
                 'admin_pool'    => $this->container->get('sonata.admin.pool'),
+                'reset_route'   => $resetRoute, // TODO: Deprecated in 2.3, to be removed in 3.0
             ));
     }
 
--- a/Resources/views/Admin/Security/login.html.twig
+++ b/Resources/views/Admin/Security/login.html.twig
@@ -57,7 +57,8 @@
                     </div>
                 </form>
 
-                <a href="{{ path('sonata_user_admin_resetting_request') }}">{{ 'forgotten_password'|trans({}, 'SonataUserBundle') }}</a>
+                {#<a href="{{ path('sonata_user_admin_resetting_request') }}">{{ 'forgotten_password'|trans({}, 'SonataUserBundle') }}</a>#}
+                <a href="{{ path(reset_route) }}">{{ 'forgotten_password'|trans({}, 'SonataUserBundle') }}</a>
             {% endblock %}
         </div>
     </div>


Controller/AdminSecurityController.php
```diff
@@ -64,12 +64,21 @@ public function loginAction()
             return new RedirectResponse($refererUri && $refererUri != $request->getUri() ? $refererUri : $this->container->get('router')->generate('sonata_admin_dashboard'));
         }
 
+        // TODO: To be removed when the major version is changed
+        $resetRoute = 'sonata_user_admin_resetting_request';
+
+        // TODO: To be removed when the major version is changed
```
Soullivaneuh: You should add a `@trigger_error` for deprecation and specify sincer version deprecation and in which version it should be removed on both comments and triggers.

Soullivaneuh:
Please next time, put refs on PR description. Keep titles and commits descriptions short and clear.

sagikazarmark:
Sorry, I put it in a commit msg, PR gets the same title automatically.

sagikazarmark:
Added trigger_error.

sagikazarmark:
Sorry for the trouble caused.

sagikazarmark:
Should I squash?

Controller/AdminSecurityController.php
Soullivaneuh: Please change on comment also.

Soullivaneuh:
Yes, after needed corrections.

Controller/AdminSecurityController.php
```diff
@@ -64,12 +64,22 @@ public function loginAction()
...
+        if (null === $this->container->get('router')->getRouteCollection()->get('sonata_user_admin_resetting_request')) {
+            @trigger_error('Using the route fos_user_resetting_request for admin password resetting is deprecated since version 2.4 and will be removed in 3.0. Use sonata_user_admin_resetting_request instead.', E_USER_DEPRECATED);
```
Soullivaneuh: Next minor of user-bundle will be 2.3, not 2.4.

sagikazarmark:
Squashed

Soullivaneuh:
Thank you @sagikazarmark.

sagikazarmark:
This is how it should regularly be solved. However I can imagine a few cases when we have to introduce such changes. Maybe we could have a twig function to handle this:

`{{ bc_path('new_path', 'orig_path', 'deprectated_version', 'remove_version') }}`

Soullivaneuh:
That is a nice idea.

sagikazarmark:
I think this should go into the CoreBundle in the form of a BcExtension for twig.

Controller/AdminSecurityController.php
```diff
@@ -64,12 +64,22 @@ public function loginAction()
             return new RedirectResponse($refererUri && $refererUri != $request->getUri() ? $refererUri : $this->container->get('router')->generate('sonata_admin_dashboard'));
         }
 
+        // TODO: Deprecated in 2.3, to be removed in 3.0
+        $resetRoute = 'sonata_user_admin_resetting_request';
+
+        // TODO: Deprecated in 2.3, to be removed in 3.0
+        if (null === $this->container->get('router')->getRouteCollection()->get('sonata_user_admin_resetting_request')) {
```
rande: This is a very expensive operation as it load route definition from files: yaml, xml, etc ...

Soullivaneuh: What is your suggestion?

sagikazarmark: The alternative would be catching a RouteNotFoundException.

rande: Exception is a better option.

Soullivaneuh: @sagikazarmark Could you please provide a PR about it?

BTW, new tests about that would be great. :+1:

sagikazarmark: Will submit the BcExtension with using Exception then. Then this solution can be replaced.

sagikazarmark: Is that OK for you? Or should I provide a PR here as well? IMO a BcExtension would be better, because we can catch the exception right at generation time.

Soullivaneuh: Don't know, wait for @rande opinion.

In any-case, this option should be done on core on something else more generic.

sagikazarmark: Will submit a BcExtension to the core anyway.

rande: There is no need for a BcExtension. Or I might missed something

sagikazarmark: See my comment in the PR above. I mentioned a possible general solution for such cases: a twig extension, with bc_path function.

sagikazarmark:
@rande something like this:

``` php
<?php

/*
 * This file is part of the Sonata package.
 *
 * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Sonata\CoreBundle\Twig\Extension;

use Symfony\Component\Routing\Exception\RouteNotFoundException;

/**
 * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
 */
class BcExtension extends \Twig_Extension
{
    private $urlGenerator;

    /**
     * @param UrlGeneratorInterface $urlGenerator
     */
    public function __construct(UrlGeneratorInterface $urlGenerator)
    {
        $this->urlGenerator = $urlGenerator;
    }

    /**
     * {@inheritdoc}
     */
    public function getFunctions()
    {
        return array(
            new \Twig_SimpleFunction('bc_path', array($this, 'getBcPath'), array('is_safe_callback' => array($this, 'isUrlGenerationSafe'))),
        );
    }

    /**
     * @param string     $name
     * @param string     $oldName
     * @param string     $deprecatedSince
     * @param string     $removeIn
     * @param array      $parameters
     * @param bool|false $relative
     *
     * @return string
     */
    public function getBcPath(
        $name,
        $oldName,
        $deprecatedSince,
        $removeIn,
        $parameters = array(),
        $relative = false
    ) {
        $relative ? UrlGeneratorInterface::RELATIVE_PATH : UrlGeneratorInterface::ABSOLUTE_PATH;

        try {
            return $this->urlGenerator->generate($name, $parameters, $relative);
        } catch (RouteNotFoundException $e) {
            @trigger_error(sprintf('Using the route %s is deprecated since version %s and will be removed in %s. Use %s instead.', $oldName, $deprecatedSince, $removeIn, $name), E_USER_DEPRECATED);

            return $this->urlGenerator->generate($oldName, $parameters, $relative);
        }
    }

    /**
     * {@inheritdoc}
     */
    public function getName()
    {
        return 'bc';
    }
}
```

sagikazarmark:
We could make this a general BC extension or specific to routes. In this case the extension could be simpler as the BcRouteExtension could extend the RoutingExtension from the Twig Bridge. Lots of code could be inherited that way.

Controller/AdminSecurityController.php
rande: not sure we need such extension for now. I will prefer just a try/catch statement with a deprecated notice.

sagikazarmark: Did it in #631

PR #630, opened 2015-09-29
Fix cyclic calling configureSettings

sajfi:
Fix cyclic calling ProfileMenuBlockService::configureSettings on block service.


OskarStark:
Does it fix #632 @sajfi ?

OskarStark:
Refs #628

Soullivaneuh:
Thank you for the fix @sajfi.

PR #640, opened 2015-11-04
Added ManagerInterface to UserManager

core23:
Maybe we should add the ``ManagerInterface`` to the ``UserManagerInterface`` and ``GroupManagerInterface`` with the next major release.

### DIFF
--- a/Entity/UserManager.php
+++ b/Entity/UserManager.php
@@ -12,6 +12,7 @@
 namespace Sonata\UserBundle\Entity;
 
 use FOS\UserBundle\Doctrine\UserManager as BaseUserManager;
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Can you please explain why?

core23:
Because all methods from the ``ManagerInterface`` were implemented and refered to a ``@inheritdoc`` which does not exist at the moment.

Soullivaneuh:
Sounds good. @rande what do you think?

core23:
ping @rande

OskarStark:
:+1:

OskarStark:
Thank you @core23

PR #642, opened 2015-11-05
Catch undefined manager_type

core23:
This could lead to an error if you use something else than orm or mongodb as an entity manager.

### DIFF
--- a/DependencyInjection/SonataUserExtension.php
+++ b/DependencyInjection/SonataUserExtension.php
@@ -176,6 +176,8 @@ class SonataUserExtension extends Extension
             $modelType = 'Entity';
         } elseif ('mongodb' === $config['manager_type']) {
             $modelType = 'Document';
+        } else {
+            throw new \InvalidArgumentException(sprintf('Invalid manager type "%s".', $config['manager_type']));
         }
 
         $defaultConfig['class']['user']  = sprintf('Application\\Sonata\\UserBundle\\%s\\User', $modelType);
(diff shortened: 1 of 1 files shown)


OskarStark:
ping @rande

core23:
@rande Should we ignore this (return) or throw an exception?

OskarStark:
i would throw an exception

ping @rande

OskarStark:
any news on this?

core23:
IMO a simple return should do it. Today there is no possibility to run this line: https://github.com/sonata-project/SonataUserBundle/blob/master/DependencyInjection/Configuration.php#L63

OskarStark:
ok :+1:

core23:
After thinking a few minutes... An exception would be better, as the other bundles do the same @OskarStark

OskarStark:
:+1:

DependencyInjection/SonataUserExtension.php
```diff
@@ -199,6 +201,8 @@ public function configureClass($config, ContainerBuilder $container)
             $modelType = 'entity';
         } elseif ('mongodb' === $config['manager_type']) {
             $modelType = 'document';
+        } else {
+            throw new \InvalidArgumentException(sprintf('SonataUserBundle - Invalid manager type "%s".', $config['manager_type']));
```
OskarStark: Do we really need the "SonataUserBundle"-prefix?

core23:
Fixed @OskarStark

OskarStark:
:+1: thank you @core23

OskarStark:
@core23 can you merge this in the other branches as well please?

Issue #450, opened 2014-07-30
Sonata\UserBundle\Document\BaseUser extends FOSUserBundle user entity


akarmazyn:
Right now Sonata\UserBundle\Document\BaseUser extends FOS\UserBundle\Entity\User (throught Sonata\UserBundle\Model\User) resulting in "No mapping found for field 'xxx'..." issue.
It should extend the FOS\UserBundle\Document\User class.

PR #652, opened 2015-11-27
Fix MongoDB Support

Id4v:
As stated in #450 MongoDB support is not working properly.

`Sonata\Document\BaseUser` extends `FOSUser\Entity\User` via the extends of `Sonata\Model\User`.

This causes a severe bug in the MongoDB Support (All FosUser fields are missing).

I chose to delete the `Model\User` class and duplicate the fields in both `Document\BaseUser` and `Entity\BaseUser`. 
Not sure this is the best solution, but this is the only one I can think of, as we cannot use Traits for PHP5.3 support.

In addition, the `SonataUserExtension` was loading the `serializer.xml` file no matter what manager was used.
However the service declared in this file depends on an "orm" service, causing an "Service not found" error.
Now it is loaded only if the manager is orm.


OskarStark:
ping @rande

Id4v:
Any thoughts about this?
Some things to improve maybe?

Soullivaneuh:

Regards.

PR #668, opened 2015-12-24
added translation for user_block_profile key in ca

mrcmorales:


OskarStark:
Thank you @mrcmorales

PR #671, opened 2016-01-02
Fix: Empty route after registration

core23:
Sometimes the route could be empty after registration. This could happen when no "sonata_user_redirect_url" value exists in the user session.


OskarStark:
Thank oyu @core23

PR #673, opened 2016-01-08
Travis: Only run supported php versions

core23:
Drop support for outdated php versions: https://secure.php.net/supported-versions.php


core23:
@rande Should we add php >=5.5 as a composer dependency?

OskarStark:
what about keeping the travis stuff, allow failures on outdated versions and add 5.5. to composer.json ?

core23:
Good idea @OskarStark 
Moved the PHP 5.3 and 5.4 stuff to the allow failure section.

OskarStark:
Thank you @core23

PR #677, opened 2016-01-15
Improve compatibility with FOSUserBundle 2.x controllers

javer:


Controller/AdminResettingController.php
```diff
@@ -27,108 +31,127 @@ class AdminResettingController extends ResettingController
...
+            'base_template' => $this->get('sonata.admin.pool')->getTemplate('layout'),
+            'admin_pool'    => $this->get('sonata.admin.pool'),
         ));
     }
 
     /**
      * {@inheritdoc}
      */
-    public function sendEmailAction()
+    public function sendEmailAction(Request $request)
```
core23: Adding/Changing method parameters is a BC break

javer: Parent class ```FOS\UserBundle\Controller\ResettingController``` from FOSUser2 has this [signature](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Controller/ResettingController.php#L43): ```public function sendEmailAction(Request $request)```.

Without this fix we get the following error:
> Runtime Notice: Declaration of Sonata\UserBundle\Controller\AdminResettingController::sendEmailAction() should be compatible with FOS\UserBundle\Controller\ResettingController::sendEmailAction(Symfony\Component\HttpFoundation\Request $request)

Please note that this PR is for branch add_support_for_fos_user2 which is entirely a BC break.

OskarStark:
@rande its your branch, please merge this if you need it

rande:
@OskarStark yes it is

PR #682, opened 2016-01-21
Fixed German and Ukrainian translations

themao:
As of recent testing of one of the projects, translation mistakes were found. Here is a proposition to fix it


OskarStark:
Thank you @themao

PR #688, opened 2016-02-04
[FOS2] Some fixes in imports

nchervyakov:
The imports of certain classes were missing in admin controllers. Added them.


nchervyakov:
Sorry for commit f016322. I do not know why it is in pull request, apparently it's because of my error. 
I only meant changes from commit 'Fix admin controller imports'

PR #690, opened 2016-02-10
Fixed build

### LINKED ISSUE #684: Documentation is broken
Build is not passing, because something is wrong with sphinx

see sphinx-doc/sphinx#2264

greg0ire:
Fixes #684

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,11 +34,11 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
     - php: 5.5
-      env: SYMFONY_VERSION="3.0.*"
+      env: SYMFONY_VERSION=3.0.*
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.*"
+      env: SYMFONY_VERSION=3.0.*
     - php: 7.0
-      env: SYMFONY_VERSION="3.0.*"
+      env: SYMFONY_VERSION=3.0.*
 
   allow_failures:
     - php: hhvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,11 +34,11 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
     - php: 5.5
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION="3.0.*"
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION="3.0.*"
     - php: 7.0
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION="3.0.*"
 
   allow_failures:
     - php: hhvm
@@ -53,7 +53,6 @@ before_script:
   - composer global require phpunit/phpunit:@stable fabpot/php-cs-fixer --no-update
   - composer global update --prefer-dist --no-interaction
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
-  - if [ "$ACL_VERSION" != "" ]; then composer require "symfony/security-acl:${ACL_VERSION}" --no-update; fi;
   - travis_wait composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
   - export PATH=$HOME/.local/bin:$PATH
   - pip install -r Resources/doc/requirements.txt --user `whoami`
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,6 @@ matrix:
       env: SYMFONY_VERSION=2.7.*
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
-    - php: 5.5
-      env: SYMFONY_VERSION=3.0.*
-    - php: 5.6
-      env: SYMFONY_VERSION=3.0.*
-    - php: 7.0
-      env: SYMFONY_VERSION=3.0.*
     - php: nightly
 
   allow_failures:
@@ -46,6 +40,7 @@ matrix:
     - php: 5.3
     - php: 5.4
     - php: nightly
+    - env: SYMFONY_VERSION="3.0.*"
 
 before_script:
   - (phpenv config-rm xdebug.ini || exit 0)
@@ -54,6 +49,13 @@ before_script:
   - composer config -q -g github-oauth.github.com $GITHUB_OAUTH_TOKEN
   - composer global require phpunit/phpunit:@stable fabpot/php-cs-fixer --no-update
   - composer global update --prefer-dist --no-interaction
+  - |
+      if [ "$SYMFONY_VERSION" == "3.0.*" ]; then
+      composer require "symfony/http-foundation" "^3.0" --no-update
+      composer require "symfony/form" "^3.0" --no-update
+      composer require "symfony/security" "^3.0" --no-update
+      composer require "symfony/console" "^3.0" --no-update
+      fi
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
   - travis_wait composer update --prefer-dist --no-interaction $COMPOSER_FLAGS
   - export PATH=$HOME/.local/bin:$PATH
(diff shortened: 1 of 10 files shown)


OskarStark:
can you add line breaks? IMO its more readable

greg0ire:
Sure!

OskarStark:
@greg0ire can you please have a look why travis fails?

greg0ire:
There are several problems, and I won't be able to work on them very soon.

One of them is about ACLs again, not sure if they are really required, I'll try to remove them.

greg0ire:
I don't think there is a stable release of FOSUserBundle compatible with 3.0. I'll move the 3.0 builds for to the allowed_failures.

greg0ire:
The 5.6 build fails with this message : 

> PHP Fatal error:  Call to undefined method Mock_Form_86db2601::bind() in /home/travis/build/sonata-project/SonataUserBundle/Controller/Api/GroupController.php on line 196

greg0ire:
The last issue, with the security context class and interface being removed is harder than expected, I won't be able to fix it soon. Should I remove this particular build from the build matrix? I'm afraid if I allow 5.6 to fail, all builds with 5.6 will be allowed to fail.

.travis.yml
```diff
@@ -34,16 +34,17 @@ matrix:
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
     - php: 5.5
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION="3.0.*"
```
core23: You can remove the quotes too

OskarStark:
allow failures for 5.6 is not an option IMO

greg0ire:
@OskarStark : can you have a look at the 5.6 build though ? I'm under the impression that some part of symfony packages is in 3.0, and the other part in < 3.0

greg0ire:
@OskarStark : the build won't run for 5.6 unless we restrict symfony to 2.8, which corresponds to another build that already exists. In other words, 5.6 will not pass until 5.6 + sf 3.0 passes :/ What should I do?

Tests/Form/Type/SecurityRolesTypeTest.php
```diff
@@ -68,12 +69,26 @@ public function testGetName()
     public function testGetParent()
     {
         $type = new SecurityRolesType($this->roleBuilder);
-        $this->assertEquals('choice', $type->getParent());
+        $this->assertEquals(
+            version_compare(Kernel::VERSION, '2.8.0', '>=') ?
```
Soullivaneuh: Don't use kernel version compare for this.

Tests/Form/Type/SecurityRolesTypeTest.php
```diff
@@ -68,12 +69,26 @@ public function testGetName()
...
+                'Symfony\Component\Form\Extension\Core\Type\ChoiceType' :
+                'choice',
+            $type->getParent()
+        );
+    }
+
+    private function getSecurityRolesTypeName()
+    {
+        if (version_compare(Kernel::VERSION, '2.8.0', '>=')) {
```
Soullivaneuh: Same as https://github.com/sonata-project/SonataUserBundle/pull/690/files#r52998441

Tests/Form/Type/SecurityRolesTypeTest.php
greg0ire: Right, I'll use the same trick as in the other P.R. (with `setDefaultOptions`)

Form/Type/SecurityRolesType.php
```diff
@@ -116,7 +117,11 @@ public function configureOptions(OptionsResolver $resolver)
      */
     public function getParent()
     {
-        return 'choice';
+        if (version_compare(Kernel::VERSION, '2.8.0', '>=')) {
```
Soullivaneuh: https://github.com/sonata-project/SonataUserBundle/pull/690/files#r52998441

.travis.yml
```diff
@@ -34,16 +34,17 @@ matrix:
...
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION=3.0.*
     - php: 7.0
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION=3.0.*
 
   allow_failures:
     - php: hhvm
     - php: 5.3
```
Soullivaneuh: Can someone explain me why the thing PHP 5.3 and 5.4 are on `allowed_failures`? O_O

BTW, test passes on those versions.

Soullivaneuh:
I think we should add `conflicts` section for Symfony 3 and leave travis clear (PHP versions without SF version, and only one SF 3 build on `allowed_failures`).

Then, work on another PR for SF 3 compatibility.

Maybe a conflict like `"symfony/symfony": ">=3.0"` will works but not sure.

.travis.yml
greg0ire: Comes from b45ef9ce76ce22ab6c18c4ad927a8d0c9f261c4c by @core23 …

greg0ire:
Let's give it a try

greg0ire:
`conflict` does not seem to work as expected, probably because we are not really requiring `symfony/symfony` here

greg0ire:
@Soullivaneuh : wouldn't it be better to simply remove the "|^3.0" for now?

Soullivaneuh:
> `conflict` does not seem to work as expected, probably because we are not really requiring `symfony/symfony` here

That's too bad that composer does not handle reverse resolution for conflicts.

> @Soullivaneuh : wouldn't it be better to simply remove the `|^3.0` version constraint parts for now?

Maybe, but you'll have to change each dependencies on Travis to test 3.0.

greg0ire:
> Maybe, but you'll have to change each dependencies on Travis to test 3.0.

Right, I'll do that.

greg0ire:
It looks better. I'll try to replace the `version_compare` calls soon.

.travis.yml
```diff
@@ -53,7 +50,14 @@ before_script:
   - composer global require phpunit/phpunit:@stable fabpot/php-cs-fixer --no-update
   - composer global update --prefer-dist --no-interaction
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require "symfony/symfony:${SYMFONY_VERSION}" --no-update; fi;
-  - if [ "$ACL_VERSION" != "" ]; then composer require "symfony/security-acl:${ACL_VERSION}" --no-update; fi;
+  - |
+      if [ "$SYMFONY_VERSION" == "3.0.*" ]
```
Soullivaneuh: `if; then` it more used AFAIK.

greg0ire: Ok changing that.

greg0ire:
@Soullivaneuh : replaced the `version_compare` calls, what do you think?

Form/Type/SecurityRolesType.php
```diff
@@ -116,7 +116,10 @@ public function configureOptions(OptionsResolver $resolver)
      */
     public function getParent()
     {
-        return 'choice';
```
Soullivaneuh: Please a comment with which SF version we are supporting here.

Soullivaneuh:
Just one asking and we are good to merge IMO. :+1:

greg0ire:
I'm on it!

Soullivaneuh:
And OFC, squash your commits. ;-)

greg0ire:
Do you know [my feelings about that](https://github.com/sonata-project/SonataPageBundle/pull/608#issuecomment-179477513)? I think I might be able to squash some of the commits, but not all of them.

Soullivaneuh:
> Do you know my feelings about that? I think I might be able to squash some of the commits, but not all of them.

I globally agree for multiple commits if each commit is related to **only** one work.

For me, it should be one PR by work and this special case of PR "To make Travis green" should never happen.

This happened this time, but let's try to not have it again. ;-)

greg0ire:
> this special case of PR "To make Travis green" should never happen.

There! That's why there are so many commits. Also, the PR contains some micro-work, that does not warrant a PR, but remains unrelated, like f6017f8 or 906a01b . It does not make the build green, but making a P.R. for that is well… a bit much.

greg0ire:
I will try to squash 543eb17 and ab73141 though

greg0ire:
Squashed the related commits, the tests are green, is this good to merge?

OskarStark:
Thank you @greg0ire

PR #692, opened 2016-02-19
[FOS2] Fix renamed service fos_user.entity_manager

nlzet:
Like mentioned in https://github.com/sonata-project/SonataUserBundle/issues/322#issuecomment-186171692 this service has been renamed in fos 2, and therefore I suppose it should be changed.

Or do we need to support both fos versions?


OskarStark:
no we only support the new 2.0 version

OskarStark:
@rande please fell free to merge this PR

PR #697, opened 2016-02-25
Updated Finnish translations

vviippola:
Updated the missing Finnish translations


vviippola:
Sorry messed up with this, two different fixes included.

PR #708, opened 2016-03-19
Update base_login.html.twig

Soolan:
if you don't mention the block name you will get an "unexpected render tag" error


core23:
* Please update the commit message and PR title to say what you did
* Squash your commits

Issue #621, opened 2015-09-23
translation not working


OskarStark:
Here we use this translation in ```de```:

```xml
            <trans-unit id="Bad credentials">
                <source>Bad credentials.</source>
                <target>Benutzername oder Passwort ungültig</target>
            </trans-unit>
```
Ref: https://github.com/sonata-project/SonataUserBundle/blob/master/Resources/translations/SonataUserBundle.de.xliff#L497

This isn't working, because of the used translation_domain (FOSUserBundle) here:
```
            {% block sonata_user_login_error %}
                {% if error %}
                    <div class="alert alert-danger">{{ error|trans({}, 'FOSUserBundle') }}</div>
                {% endif %}
            {% endblock %}
```

but how to fix this problem, i created a ```FOSUserBundle.de.xliff``` file in ```app/Resources/translations```, but how to fix this for this bundle?

Oskar

ejkun:
I was looking in FOSUserBundle latest commits about this. They took out the `Bad Credentials.` message, changed the `SecurityController` and changed the `login.html.twig` file, all in this [commit](https://github.com/FriendsOfSymfony/FOSUserBundle/commit/848884313726c81a42f86c0952957872d3c532e2). 

All of this should be fixed with the FOSUserBundle 2.0 version. But IMO, it will take a while. 
I managed to fix the error doing the same changes they made. 
But with the controller, we will have to copy they function and change the lines about the `$error` variable. I don't have time to do it right now, but later today or tomorrow I will make the changes in my fork and see if it pass all the tests.

ejkun:
As I said in the last comment, I found a way to fix it. But since there were many changes in the repository structure, the PR was closed, but the problem still there. @OskarStark , should I rebuild the PR for the 3.x branch? Beacuse it's just a quickfix, if a stable 2.x version of FOSUserBundle is released, everything done in the PR will have to be removed, since the controller itself would stop calling FOSUserBundle controller. What do you think?

PR #716, opened 2016-04-08
Fixed the 'Bad Credentials' translation problem

ejkun:
As related in this [issue](https://github.com/sonata-project/SonataUserBundle/issues/621) there was a problem with the error message in the login page, that was not being translated. That was a error with FOSUserBundle, that apparently will be fixed in the 2.0 version, but since there is no prediction to when it will be ready. I made some changes by myself based on the dev-master of FOSUserBundle.

I changed both security controllers and the templates for the login page to look more like the latest version of FOSUserBundle. Now the error message is translated with the 'security'. Also took out the Bad Credentials entry in one of the translation files since there is no need for it.

### DIFF
--- a/Controller/SecurityFOSUser1Controller.php
+++ b/Controller/SecurityFOSUser1Controller.php
@@ -37,6 +37,36 @@ class SecurityFOSUser1Controller extends SecurityController
             return new RedirectResponse($url);
         }
 
-        return parent::loginAction();
+        /*
+            Implementation of parent::loginAction
+            Needed for fixing the "Bad Credentials" translation problem
+            The code is a mix of the 2.0.0 version
+            and the 1.3.6 version of FOSUserBundle
+        */
+
+        $request = $this->container->get('request');
+        /* @var $request \Symfony\Component\HttpFoundation\Request */
+        $session = $request->getSession();
+        /* @var $session \Symfony\Component\HttpFoundation\Session\Session */
+
+        // get the error if any (works with forward and redirect -- see below)
+        if ($request->attributes->has(SecurityContextInterface::AUTHENTICATION_ERROR)) {
+            $error = $request->attributes->get(SecurityContextInterface::AUTHENTICATION_ERROR);
+        } elseif (null !== $session && $session->has(SecurityContextInterface::AUTHENTICATION_ERROR)) {
+            $error = $session->get(SecurityContextInterface::AUTHENTICATION_ERROR);
+            $session->remove(SecurityContextInterface::AUTHENTICATION_ERROR);
+        } else {
+            $error = null;
+        }
+
+        if (!$error instanceof AuthenticationException) {
+            $error = null; // The value does not come from the security component.
+        }
+
+        return $this->renderLogin(array(
+            'last_username' => (null === $session) ? '' : $session->get(SecurityContextInterface::LAST_USERNAME),
+            'error'         => $error,
+            'csrf_token'    => $this->container->get('form.csrf_provider')->generateCsrfToken('authenticate'),
+        ));
     }
 }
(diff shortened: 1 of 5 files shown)


OskarStark:
nice job @ejkun 

can you please add a test case for this implementation?

ejkun:
@OskarStark , I'm not very good with tests. Which test should I do? Test the controller? Since there are no tests for the Controllers or the Views, I don't know where to start.

ejkun:
I was checking the FOSUserBundle to see if they have some tests. Well, they don't. None of the controllers or views are tested. So I really don't know where to start.

greg0ire:
@OskarStark I don't think asking for tests is fair in that case.

greg0ire:
@ejkun , about your commit message, can you read and apply : https://github.com/Soullivaneuh/sonata-dev-kit/blob/doc-contributing/doc/CONTRIBUTING.md#the-commit-message (especially the fix about part).

Also, your commit message descriptions contains the word "also". There should probably be two commits. Use

```
git reset origin/master
git add --patch
```
to forget your commit and rebuild them block by block.

Resources/views/Admin/Security/login.html.twig
```diff
@@ -30,7 +30,7 @@
             {% block sonata_user_login_form %}
                 {% block sonata_user_login_error %}
                     {% if error %}
-                        <div class="alert alert-danger">{{ error|trans({}, 'FOSUserBundle') }}</div>
+                        <div class="alert alert-danger alert-error">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
```
greg0ire: This line is too long.

Resources/views/Security/base_login.html.twig
```diff
@@ -25,7 +25,7 @@ file that was distributed with this source code.
 
                     {% block sonata_user_login_error %}
                         {% if error %}
-                            <div class="alert alert-danger alert-error">{{ error|trans({}, 'FOSUserBundle') }}</div>
+                            <div class="alert alert-danger alert-error">{{ error.messageKey|trans(error.messageData, 'security') }}</div>
```
greg0ire: This line is too long.

Controller/SecurityFOSUser1Controller.php
```diff
@@ -37,6 +37,41 @@ public function loginAction()
...
+        }
+
+        if (!$error instanceof AuthenticationException) {
+            $error = null; // The value does not come from the security component.
+        }
+
+        // last username entered by the user
+        $lastUsername = (null === $session) ? '' : $session->get(SecurityContextInterface::LAST_USERNAME);
+
+        $csrfToken = $this->container->get('form.csrf_provider')->generateCsrfToken('authenticate');
```
greg0ire: Both these variables are used only once, as values for keys that have the same name. You could write directly :

```php
array(
    'last_username' => (null === $session) ? '' : $session->get(SecurityContextInterface::LAST_USERNAME);
    'csrf_token' => $this->container->get('form.csrf_provider')->generateCsrfToken('authenticate');
)
```

Controller/SecurityFOSUser1Controller.php
```diff
@@ -37,6 +37,41 @@ public function loginAction()
             return new RedirectResponse($url);
         }
 
-        return parent::loginAction();
+        /*
+            Implementation of parent::loginAction
+            Needed for fixing the "Bad Credentails" translation problem
+            The code is a mix of current dev-master
```
greg0ire: "current" will lose its meaning ;)

Controller/SecurityFOSUser1Controller.php
```diff
@@ -37,6 +37,41 @@ public function loginAction()
             return new RedirectResponse($url);
         }
 
-        return parent::loginAction();
+        /*
+            Implementation of parent::loginAction
+            Needed for fixing the "Bad Credentails" translation problem
```
greg0ire: Credentials

ejkun:
ping @OskarStark

Soullivaneuh:

Regards.

PR #722, opened 2016-04-21
Update version number for admin bundle

dragosprotung:
Version 2.4 does not exist anymore.
See https://github.com/sonata-project/SonataAdminBundle/issues/3731 for details.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "symfony/security-acl": "~2.8|~3.0",
         "symfony/console": "~2.8|~3.0",
         "sonata-project/core-bundle": "~2.3,>=2.3.10",
-        "sonata-project/admin-bundle": "~2.3|~2.4@dev",
+        "sonata-project/admin-bundle": "~2.3|~3.0@dev",
         "friendsofsymfony/user-bundle": "~2.0@dev",
         "sonata-project/doctrine-extensions": "~1.0",
         "sonata-project/easy-extends-bundle": "~2.1",


OskarStark:
@Soullivaneuh can you please have a look?

Soullivaneuh:
3.0 bundles are WIP. You can follow work here: https://github.com/sonata-project/SonataAdminBundle/issues/3731

Sonata user it's a bit more complex than user. I'll have to work on it with @th3mouk and @rande.

If you want to get it working this, use aliases on you composer.json like described here: https://github.com/sonata-project/SonataAdminBundle/issues/3731#issuecomment-212636270

Closing this one.

dragosprotung:
@Soullivaneuh The only reason for this PR is because 2.4 does not exist anymore in the admin bundle, so composer will fail to install the user bundle even if i do not use the admin bundle

Soullivaneuh:
@dragosprotung You can use composer alias waiting for the fix: https://getcomposer.org/doc/articles/aliases.md#require-inline-alias

sagikazarmark:
I am not entirely sure about the validity of the reasoning, given that using `~3.0@dev` is valid in other packages?

https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/master/composer.json#L28

Not complaining, I am perfectly fine with aliasing.

Soullivaneuh:
@sagikazarmark This is WIP and sonata-user is a bit more complicated because of FOSUser 2.0 WIP too.

We will look at this ASAP.

sagikazarmark:
Understood.

PR #726, opened 2016-05-08
Update composer.json

TheRatG:
Change admin bundle version, because composer cannot find 2.4@dev.

I have to used some trick now in my composer
```
...
"sonata-project/admin-bundle": "3.x-dev as 2.4.x-dev",
"sonata-project/user-bundle": "2.3.x-dev",
...
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "symfony/security": "~2.3",
         "symfony/console": "~2.3",
         "sonata-project/core-bundle": "~2.3,>=2.3.10",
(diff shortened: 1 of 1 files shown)


Soullivaneuh:
Thanks for your contribution, but please for now use branch aliasing on composer, and don't try to update it from PR.

See: https://github.com/sonata-project/SonataUserBundle/pull/722#issuecomment-213022133

TheRatG:
Ok, I will not. Thanks for your answer.

PR #720, opened 2016-04-20
simplified parameter handling

OskarStark:
IMO this is much cleaner

PR #727, opened 2016-05-10
Simplified parameter handling

OskarStark:
IMO this is much cleaner!

i added some more tests in a separate commit

This PR replaces #720, because it wasn't done on a forked repo, sorry!

### DIFF
--- a/Tests/Entity/UserManagerTest.php
+++ b/Tests/Entity/UserManagerTest.php
@@ -42,7 +42,8 @@ class UserManagerTest extends \PHPUnit_Framework_TestCase
                     $self->equalTo('u.username'),
                     $self->equalTo('ASC')
                 );
-                $qb->expects($self->once())->method('setParameters')->with($self->equalTo(array()));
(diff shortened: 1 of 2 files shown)


OskarStark:
ping @greg0ire @Soullivaneuh

Tests/Entity/UserManagerTest.php
```diff
@@ -42,7 +42,8 @@ public function testGetPager()
                     $self->equalTo('u.username'),
                     $self->equalTo('ASC')
                 );
-                $qb->expects($self->once())->method('setParameters')->with($self->equalTo(array()));
+                $qb->expects($self->never())->method('setParameter');
+                $qb->expects($self->never())->method('setParameters');
```
OskarStark: not sure if this is needed

greg0ire: We should test how the object behaves, not how it is implemented.

OskarStark: what do you mean by that?

PR #653, opened 2015-11-28
Make google-authenticator dependency optional

core23:
### Changelog

```markdown
### Changed
- Made `sonata-project/google-authenticator` an optional dependency
```

### Subject

The google-authenticator is an optional dependency now

### To do

- [X] Update composer file
- [X] Update docs

PR #739, opened 2016-05-19
Make Google Authenticator dependency optional

sagikazarmark:
Resubmission of sonata-project/SonataUserBundle#653

### Changelog

```markdown
### Changed
- Google Authenticator is now optional, you have to install it manually.
```

### Subject

This PR removes the mandatory Google Authenticator dependency. It provides a feature, not a core functionality, so making it optional actually makes sense.


Soullivaneuh:
We have the same work in #653 as a major one.

I think it should be a major because we remove a possibly used dependency.

OskarStark:
Closing in favor of #653

PR #740, opened 2016-05-19
Remove deprecated Admin class, use AbstractAdmin

sagikazarmark:
### Changelog

```markdown
### Changed
- Admin classes extend `Sonata\AdminBundle\Admin\AbstractAdmin`
```

### Subject

Removes deprecated Admin class usage.

### DIFF
--- a/Admin/Model/GroupAdmin.php
+++ b/Admin/Model/GroupAdmin.php
@@ -11,12 +11,12 @@
 
 namespace Sonata\UserBundle\Admin\Model;
 
-use Sonata\AdminBundle\Admin\Admin;
+use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Form\FormMapper;
 
-class GroupAdmin extends Admin
+class GroupAdmin extends AbstractAdmin
 {
     /**
      * {@inheritdoc}
--- a/Admin/Model/UserAdmin.php
+++ b/Admin/Model/UserAdmin.php
@@ -12,13 +12,13 @@
 namespace Sonata\UserBundle\Admin\Model;
 
 use FOS\UserBundle\Model\UserManagerInterface;
-use Sonata\AdminBundle\Admin\Admin;
+use Sonata\AdminBundle\Admin\AbstractAdmin;
 use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Show\ShowMapper;
 
-class UserAdmin extends Admin
+class UserAdmin extends AbstractAdmin
 {
     /**
      * @var UserManagerInterface
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "symfony/security": "^2.3",
         "symfony/console": "^2.3",
         "sonata-project/core-bundle": "^3.0",
-        "sonata-project/admin-bundle": "^3.0",
+        "sonata-project/admin-bundle": "^3.1",
         "sonata-project/datagrid-bundle": "^2.2",
         "friendsofsymfony/user-bundle": "^1.3",
         "sonata-project/doctrine-extensions": "^1.0",


Soullivaneuh:
I think this will not work with admin-bundle `^3.0`. Could you add some tests on those two classes?

sagikazarmark:
Why not?

Tests in general, or test that it doesn't work with admin 3.0?

Sullivan SENECHAL <notifications@github.com> ezt írta (2016. május 20.,
péntek):

Soullivaneuh:
Because the class renaming was done in admin-bundle 3.1, not 3.0: https://github.com/sonata-project/SonataAdminBundle/commit/b369f59375608a4c5f1b5d7fccf1ee6763dd7f9a

> Tests in general, or test that it doesn't work with admin 3.0?

For the moment, just add `GroupAdminTest` and `UserAdmin` with at minimum instanciation and if you can/want, some simple test.

sagikazarmark:
Oh, right, I missed that. What do you propose? Raise the required version?

Soullivaneuh:
Yes. Since this project follow semver, we can do this without pain.

I would like to keep this on a `patch` but I'm not sure if this is the right way. @sonata-project/contributors what do you think?

Soullivaneuh:
@sagikazarmark I'm working for simple tests.

greg0ire:
This is kind of a grey area : 
- on one hand, raising a version constraint for a dependency means the user of the lib has do something (especially if he already requires the dependency), namely upgrading the dependency
- on the other hand, it's a minor raise, which means the user should not experience any BC-break

I'd say follow your common sense. Mine tells me this should be considered minor : nothing is fixed, but a new thing is added: more forward-compatibility with 4.0

sagikazarmark:
Raised the version.

Soullivaneuh:
> nothing is fixed, but a new thing is added: more forward-compatibility with 4.0

I don't agree on the "nothing is fixed". The triggered error is a deprecated error, but it's still an error.

So yes, for me it's a fix.

BTW, in this case, we should push a minor every time we fix a deprecation even for Symfony. In this case, we will not having any patch either. :-)

My main concern is about the minor raising, and you answered it:

> it's a minor raise, which means the user should not experience any BC-break

Soullivaneuh:
@sagikazarmark You will be able to rebase your PR once #741 will be merged.

sagikazarmark:
Sure, just give me a hint. Although those changes won't conflict, so why exactly do I need to rebase?

Soullivaneuh:
@greg0ire http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

I would like you to focus on: 

> Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality.

So if I understand well, semver says that we could update dependencies on a patch (for a fix) if this does not heart the package API.

This is the case so I think this should be kept as a patch.

What do you think?

Soullivaneuh:
> Sure, just give me a hint. Although those changes won't conflict, so why exactly do I need to rebase?

You have to rebase to have my tests classes on your branch and be sure all is good. ;-)

sagikazarmark:
> This is the case so I think this should be kept as a patch.

I agree, this is not a new feature.

> You have to rebase to have my tests classes on your branch and be sure all is good. ;-)

kk

greg0ire:
You changed my mind, this is indeed a patch :)

Soullivaneuh:
@sagikazarmark Could you please rebase?

sagikazarmark:
Done

Soullivaneuh:
Thank you @sagikazarmark

webdevilopers:
Sorry for commenting this merged issue. But what are the recommended versions then?

```javascript
        "sonata-project/admin-bundle": "3.x-dev",
        "sonata-project/user-bundle": "dev-master",
```

and

```javascript
        "sonata-project/admin-bundle": "3.x-dev",
        "sonata-project/user-bundle": "3.x-dev",
```

both throw

> ClassNotFoundException in UserAdmin.php line 21:
Attempted to load class "AbstractAdmin" from namespace "Sonata\AdminBundle\Admin".
Did you forget a "use" statement for another namespace?

Soullivaneuh:
@webdevilopers `AbstractAdmin` is available since admin-bundle 3.1.

Issue #730, opened 2016-05-12
Version upgrade plan


Soullivaneuh:
I discussed with @rande about the user-bundle problematic.

### Actual state

The last stable version of user-bundle is `2.2.4`.
Several un-tagged commits are present on 2.2 branch: https://github.com/sonata-project/SonataUserBundle/compare/2.2.4...2.2

We have two branch without any stable tag:

* `2.3`
* master that is also aliased to `2.3.x`. Don't ask me why, I don't know.

For the moment, user-bundle is compatible with FOSUserBundle `1.x` stable.

### The plan

We won't get the risk to release a `2.3` stable version without being sure this one is BC with `2.2`.
Why? Simply because we will drop `2.x` support just after that (like core-bundle and admin-bundle)

Here is the steps of the elaborated plan:

- [x] Push a last version of `2.2`, `2.2.5`, to keep un-tagged work. If you was on `2.2.4` and `2.2.5` does not work, simply stay at `2.2.4`.
- [x] Push a `2.x` branch from `2.2`, for keeping the history. Set it to `2.x-dev`
- [x] Remove `2.0`, `2.1` and `2.2` branch. Push tags before, if trailing commits are present.
- [x] Merge `2.3` onto master, and remove it.
- [x] Rework composer and set it to `3.x-dev`
- [x] Prepare and release a `3.0` stable version
- [x] Prepare master for the next major (`4.x`)

### FOSUserBundle 2.0 special case

Lot of people are asking to have a support for FOSUserBundle 2.0.

@rande Tried to get it working here: https://github.com/sonata-project/SonataUserBundle/tree/add_support_for_fos_user2

But it's apparently too hard to keep both support of 1.x and 2.x.

As FOSUserBundle 2.0 is not stable yet, we **will not** add support of this one, neither for 2.x or 3.x version of our bundle.

For the future, two possible cases:

* If we can keep support for both 1.x and 2.x version of FOSUser, it will be done on next minor (probably 3.1)
* If not, the work for FOSUser 2.x support will be merged and release on a new major **only when** FOSUser 2.0 will be stable

Thanks for your support and your patience.

redactuk:
[...]

PR #745, opened 2016-05-23
[WIP] FOSUser 2.0 compatibility

Soullivaneuh:
### Changelog

``` markdown
### Added
- Now compatible with FOSUserBundle 2.0
```
### Subject

Please note this is **WIP** to **TEST** if we can make this bundle compatible with both fos-user bundle v1 and v2.

I give **NO WARRANTY** this will work. It's a try. If it succeed, we can push this on a new minor.

Even if this works, please note the FOSUser 2.x support, even on a stable release of sonata-user-bundle, will be a **best effort**. Indeed, it's hard to ensure the support of an unstable dependency.

Before you asking, this PR will be **NOT** about SF 3 compatibility but only FOSUser. The SF 3 compatibility will depends of the success of this PR, or the stabilization of FOSUser v2 and will be worked on another PR.

I don't have sonata-user-bundle installed on any maintained private projects right now.
I'll trust unit test ATM, but I'll need a lot of feedback from all of you (if you can) to test this branch on your project against both FOSUser v1 and v2 and tell me what is not working and why. Thanks for help. :+1: 
### To do
- [ ] Make it compatible with both FOSUser v1 and v2
- [ ] Add an upgrade note
- [ ] Remove Travis hack. Update dev-kit instead.

### DIFF
--- a/SonataUserBundle.php
+++ b/SonataUserBundle.php
@@ -11,6 +11,9 @@
 
 namespace Sonata\UserBundle;
 
+use Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass;
+use Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler\DoctrineMongoDBMappingsPass;
+use Doctrine\Bundle\CouchDBBundle\DependencyInjection\Compiler\DoctrineCouchDBMappingsPass;
 use Sonata\CoreBundle\Form\FormHelper;
 use Sonata\UserBundle\DependencyInjection\Compiler\GlobalVariablesCompilerPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -44,6 +47,7 @@ class SonataUserBundle extends Bundle
         $container->addCompilerPass(new GlobalVariablesCompilerPass());
 
         $this->registerFormMapping();
+        $this->addFOS1RegisterMappingsPass($container);
     }
 
     /**
@@ -73,5 +77,30 @@ class SonataUserBundle extends Bundle
             'sonata_user_api_form_group' => 'Sonata\UserBundle\Form\Type\ApiGroupType',
             'sonata_user_api_form_user' => 'Sonata\UserBundle\Form\Type\ApiUserType',
         ));
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     */
+    private function addFOS1RegisterMappingsPass(ContainerBuilder $container)
+    {
+        $fosMappingPath = realpath($container->getParameter('kernel.root_dir').'/../vendor/friendsofsymfony/user-bundle/FOS/UserBundle/Resources/config/doctrine');
+
+        if (!is_dir($fosMappingPath)) {
+            return;
+        }
+
+        $mappings = array(
+            $fosMappingPath => 'FOS\UserBundle\Model',
+        );
+        if (class_exists('Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass')) {
+            $container->addCompilerPass(DoctrineOrmMappingsPass::createXmlMappingDriver($mappings, array('fos_user.model_manager_name'), 'fos_user.backend_type_orm'));
+        }
+        if (class_exists('Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler\DoctrineMongoDBMappingsPass')) {
+            $container->addCompilerPass(DoctrineMongoDBMappingsPass::createXmlMappingDriver($mappings, array('fos_user.model_manager_name'), 'fos_user.backend_type_mongodb'));
+        }
+        if (class_exists('Doctrine\Bundle\CouchDBBundle\DependencyInjection\Compiler\DoctrineCouchDBMappingsPass')) {
+            $container->addCompilerPass(DoctrineCouchDBMappingsPass::createXmlMappingDriver($mappings, array('fos_user.model_manager_name'), 'fos_user.backend_type_couchdb'));
+        }
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ matrix:
     - php: '7.0'
       env: FOS_USER=1.3.*
     - php: '7.0'
+      env: FOS_USER=2.0.*@dev
+    - php: '7.0'
       env: FOS_USER=dev-master@dev
     - php: '7.0'
       env: SONATA_CORE=3.*
--- a/Document/BaseGroup.php
+++ b/Document/BaseGroup.php
@@ -11,7 +11,7 @@
 
 namespace Sonata\UserBundle\Document;
 
-use FOS\UserBundle\Document\Group as AbstractedGroup;
+use FOS\UserBundle\Model\Group as AbstractedGroup;
 
 /**
  * Represents a Base Group Document.
--- a/Entity/BaseGroup.php
+++ b/Entity/BaseGroup.php
@@ -11,7 +11,7 @@
 
 namespace Sonata\UserBundle\Entity;
 
-use FOS\UserBundle\Entity\Group as AbstractedGroup;
+use FOS\UserBundle\Model\Group as AbstractedGroup;
 
 /**
  * Represents a Base Group Entity.
--- a/Entity/GroupManager.php
+++ b/Entity/GroupManager.php
@@ -11,7 +11,7 @@
 
 namespace Sonata\UserBundle\Entity;
 
-use FOS\UserBundle\Entity\GroupManager as BaseGroupManager;
+use FOS\UserBundle\Doctrine\GroupManager as BaseGroupManager;
 use Sonata\DatagridBundle\Pager\Doctrine\Pager;
 use Sonata\DatagridBundle\ProxyQuery\Doctrine\ProxyQuery;
 use Sonata\UserBundle\Model\GroupManagerInterface;
--- a/Model/User.php
+++ b/Model/User.php
@@ -11,7 +11,7 @@
 
 namespace Sonata\UserBundle\Model;
 
-use FOS\UserBundle\Entity\User as AbstractedUser;
+use FOS\UserBundle\Model\User as AbstractedUser;
 
 /**
  * Represents a User model.
--- a/Tests/Controller/Api/GroupControllerTest.php
+++ b/Tests/Controller/Api/GroupControllerTest.php
@@ -55,7 +55,7 @@ class GroupControllerTest extends \PHPUnit_Framework_TestCase
         $group = $this->getMock('FOS\UserBundle\Model\GroupInterface');
 
         $groupManager = $this->getMock('Sonata\UserBundle\Model\GroupManagerInterface');
-        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('FOS\UserBundle\Entity\Group'));
+        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('Sonata\UserBundle\Entity\BaseGroup'));
         $groupManager->expects($this->once())->method('updateGroup')->will($this->returnValue($group));
 
         $form = $this->getMockBuilder('Symfony\Component\Form\Form')->disableOriginalConstructor()->getMock();
@@ -74,7 +74,7 @@ class GroupControllerTest extends \PHPUnit_Framework_TestCase
     public function testPostGroupInvalidAction()
     {
         $groupManager = $this->getMock('Sonata\UserBundle\Model\GroupManagerInterface');
-        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('FOS\UserBundle\Entity\Group'));
+        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('Sonata\UserBundle\Entity\BaseGroup'));
 
         $form = $this->getMockBuilder('Symfony\Component\Form\Form')->disableOriginalConstructor()->getMock();
         $form->expects($this->once())->method('submit');
@@ -93,7 +93,7 @@ class GroupControllerTest extends \PHPUnit_Framework_TestCase
         $group = $this->getMock('FOS\UserBundle\Model\GroupInterface');
 
         $groupManager = $this->getMock('Sonata\UserBundle\Model\GroupManagerInterface');
-        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('FOS\UserBundle\Entity\Group'));
+        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('Sonata\UserBundle\Entity\BaseGroup'));
         $groupManager->expects($this->once())->method('findGroupBy')->will($this->returnValue($group));
         $groupManager->expects($this->once())->method('updateGroup')->will($this->returnValue($group));
 
@@ -115,7 +115,7 @@ class GroupControllerTest extends \PHPUnit_Framework_TestCase
         $group = $this->getMock('FOS\UserBundle\Model\GroupInterface');
 
         $groupManager = $this->getMock('Sonata\UserBundle\Model\GroupManagerInterface');
-        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('FOS\UserBundle\Entity\Group'));
+        $groupManager->expects($this->once())->method('getClass')->will($this->returnValue('Sonata\UserBundle\Entity\BaseGroup'));
         $groupManager->expects($this->once())->method('findGroupBy')->will($this->returnValue($group));
 
         $form = $this->getMockBuilder('Symfony\Component\Form\Form')->disableOriginalConstructor()->getMock();
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "sonata-project/core-bundle": "^3.0",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/datagrid-bundle": "^2.2",
-        "friendsofsymfony/user-bundle": "^1.3",
+        "friendsofsymfony/user-bundle": "^1.3 || ^2.0@dev",
         "sonata-project/doctrine-extensions": "^1.0",
         "sonata-project/easy-extends-bundle": "^2.1",
         "sonata-project/google-authenticator": "^1.0",
@@ -35,6 +35,7 @@
     },
     "require-dev": {
         "sonata-project/seo-bundle": "^2.0",
+        "doctrine/doctrine-bundle": "^1.6",
         "doctrine/orm": "^2.0",
         "symfony/phpunit-bridge": "^2.8 || ^3.0",
         "sllh/php-cs-fixer-styleci-bridge": "^2.0"


Soullivaneuh:
Well, I fixed the test by changing the namespace for class inheritance, but I'm pretty sure it will not works because of this:

> This bundle now registers mappings for the classes in the Model namespace directly, instead of having empty extending classes.

From: https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Upgrade.md#user-and-group-class

So developer using my branch and FOSUser 1.3 will have issue with doctrine mapping.

Can anybody confirm this with more details?

In this case, we have to find a solution to have a BC inheritance. Any suggestion? The only idea I have is to have multiple classes, ones for FOS1 and others for FOS2.

Soullivaneuh:
The another solution could be to use your own [mapping registering](http://symfony.com/doc/master/cookbook/doctrine/mapping_model_classes.html).

With that, maybe can automatically register mapping for doctrine for our user/group properties **and** properties of FOSUser if the v1 is installed.

In this case, we simply have to deprecated unneeded model classes. Maybe be we should write an "Upgrade" note for this and update the install documentation, because the classes would not work with FOSUser v2.

It seems to be a good idea, but I'm saying that without know if we can make it. I have to investigate.

Soullivaneuh:
I tried with FOSUser v1. As expected, doctrine maps admin-user entities properties but not fos-user ones.

OskarStark:
> I tried with FOSUser v1. As expected, doctrine maps admin-user entities properties but not fos-user ones.

we should release 3.x with support for FOS User 1.3 and 4.x with FOS User 2.0

Soullivaneuh:
> we should release 3.x with support for FOS User 1.3 

User-bundle already have a stable release of 3.x.

> and 4.x with FOS User 2.0

I won't drop support of FOSUser v1 for v2 without know when the stable release of FOSUser v2 will be out.

See: https://github.com/sonata-project/SonataUserBundle/issues/730

OskarStark:
> without know when the stable release of FOSUser v2 will be out.

yes but waiting instead isn't the correct way i think. we should pin this or use the 2.0 alpha version. If they are using semver, they are not allowed to add BC breaks in the further beta or stable releases

Soullivaneuh:
> yes but waiting instead isn't the correct way i think. we should pin this or use the 2.0 alpha version. If they are using semver, they are not allowed to add BC breaks in the further beta or stable releases

Wrong. Alpha tags can be BC break. BTW, the last alpha tag of FOS does not allow SF 3, so the problem will be the same.

This is not a requirement IMO. FOSUser is unstable, if developer already use it, they know the risks.

Another branch compatible with FOSUser 2 is here: https://github.com/sonata-project/SonataUserBundle/tree/add_support_for_fos_user2

At least, if I can't make this PR working, we can maybe accept this configuration:

3.x: The stable branch
master: The next major branch (4.x)
master-fos-2: An additional branch, synced with master, for fos2 support.

But we wont tag a stable release depending on an unstable one.

Th3Mouk:
We don't know yet when the 4.x release will be necessary for sonata, but probably not before 2017, this let few months to help 2.0 fos/user-bundle release.
I know that we must never rely on a master branch, but our master branch in sonatauserbundle can exceptionally plug on it.
In this way we can work both on sf3 support and fos/user stabilisation, always in order to drop sf2 support for 4.0 release and fos/user 1.x.

Concerning 3.1 version of sonete user, we don't have the choice to support both fos/user 1 and 2, if we want to respect the logic that we implement since several months.

Soullivaneuh:
@OskarStark @Th3Mouk Please note one thing: Tagging a stable version user-bundle depending on an unstable version of fos-user is **not acceptable**.

So, actually two solution:

* If we can support both v1 and v2 on 3.x, let's do this.
* If we can't, we will keep FOS v1 for now.

BTW, this is not the subject of the PR. Please open another issue or comment an existing one.

OskarStark:
> @OskarStark @Th3Mouk Please note one thing: Tagging a stable version user-bundle depending on an unstable version of fos-user is not acceptable.

what about forking the repo and release a stable forked version on our own and switch later back to fos user?

Soullivaneuh:
> what about forking the repo and release a stable forked version on our own and switch later back to fos user?

Why do so complicated thing? What will we have? Two sonata-user bundle package? This will be too confusing.

If we can't support both, I'll manage to have a `master-fos-2` branch in this case. But this will be unstable and for good reasons.

OskarStark:
> If we can't support both, I'll manage to have a master-fos-2 branch in this case. But this will be unstable and for good reasons.

👍

Soullivaneuh:
With FOS v1, only the mapping of sonata is managed by doctrine, so field is missing.

I tried to backport mapping register from v2 to v1 on `addFOS1RegisterMappingsPass`, but it seems to have no effect.

Has anybody an idea?

OskarStark:
> I tried to backport mapping register from v2 to v1 on addFOS1RegisterMappingsPass, but it seems to have no effect.

so you got an error or anything else?

Soullivaneuh:
No. It's just change nothing.

Sullivan SENECHAL

OskarStark:
but the code of the compiler pass is executed? just asking

greg0ire:
@dbu, can you help? I see you are the author of `DoctrineOrmMappingsPass`… @Soullivaneuh : you're right, this author tag can be helpful :)

Soullivaneuh:
@dbu Indeed, if you have some suggestion, I would enjoy them. Currently, I'm quite stuck...

SonataUserBundle.php
```diff
@@ -74,4 +78,29 @@ public function registerFormMapping()
...
+
+        if (!is_dir($fosMappingPath)) {
+            return;
+        }
+
+        $mappings = array(
+            $fosMappingPath => 'FOS\UserBundle\Model',
+        );
+        if (class_exists('Doctrine\Bundle\DoctrineBundle\DependencyInjection\Compiler\DoctrineOrmMappingsPass')) {
+            $container->addCompilerPass(DoctrineOrmMappingsPass::createXmlMappingDriver($mappings, array('fos_user.model_manager_name'), 'fos_user.backend_type_orm'));
```
dbu: why do you need to do this? afaik the FOSUserBundle is supposed to do this itself, if you tell it to use orm mappings.

Soullivaneuh: > FOSUserBundle is supposed to do this itself

For v2/unstable version, we want to support both until v2 is not stable.

dbu:
those passes are initialized by FOSUserBundle. see https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/FOSUserBundle.php#L39-L56 and https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/DependencyInjection/FOSUserExtension.php#L56 (are you missing to configure the FOS user bundle?)

you might be seeing the effect of https://github.com/FriendsOfSymfony/FOSUserBundle/commit/08f6481c794f1d1cfb9d6eedb57b8ba08223ed08 though i think not.

SonataUserBundle.php
dbu: are you saying FOSUserBundle master does not do this? are you sure its not just missing configuration that is needed for FOSUserBundle?

greg0ire: > afaik the FOSUserBundle is supposed to do this itself

It does so only in FOS v2

Soullivaneuh:
@dbu You are talking about FOSUser v2. We want to support both of them.

dbu:
oh crap. i thought FOSUserBundle 1.3 was already using the new set up. but its not. so i guess you indeed need to instantiate the pass for user bundle 1 to be able to extend the model class directly. you maybe could do a check if FOS\UserBundle\Entity\User exist. if it does not, you don't need the compiler pass.

SonataUserBundle.php
dbu: sorry, my bad. indeed v1 relies on putting an empty class in Entity | Document to trigger mapping loading.

Soullivaneuh:
@dbu This is not really the current problem.

I can add the compiler pass with the right files. But actually, it does nothing more on doctrine mapping.

SonataUserBundle.php
dbu: i think with FOSUserBundle v1 the parameter `fos_user.backend_type_orm` is not set, so this won't trigger. pick a name that you set in this bundle.

responseconcepts:
Maybe my tweaks may be of some use: https://github.com/sonata-project/SonataUserBundle/issues/322#issuecomment-223282830

Soullivaneuh:
@responseconcepts Does your branch support **both** FOSUser v1 and v2?

responseconcepts:
@Soullivaneuh Rather not, I just tried to tweak it in a way that Symfony 3.x can use it.

TerjeBr:
Hi.
I see no work on this PR since June 2016.

Has this PR been abandoned?

What is the status of FosUserBundle 2.0 intergration now?

TerjeBr:
Another question: Since FosUserBundle now has 2.0-beta1 version (since November 2016), is there a chance that you can use that to make a Symfony 3 compitable version of the SonataUserBundle?

core23:
We won't release a stable version until there is a stable FOSUserBundle version. There is a branch which will support FOSUserBundle 2 https://github.com/sonata-project/SonataUserBundle/tree/add_support_for_fos_user2

TerjeBr:
The readme.md at https://github.com/sonata-project/SonataUserBundle/tree/add_support_for_fos_user2 tells me that it can be installed both extending FOSUserBundle and not extending the FOSUserBundle. Which option should I go for? What is the differences?

greg0ire:
This was introduced in 374b9397f28ab52de15208486a497e98b855e9c5 by @rande, with the same questions…

TerjeBr:
I read the comments on 374b939 and there was a similar question there, but no answers.

So my question still stands, what are the implications of extending or not extending from the FOSUserBundle? What are the differences?

Is it so that to be able to get advantages of the FosUserBundle 2.0 you have to extend from it, in or to not use an inferior internal implementation? Or is it the other way around, that most people should not extend from any FOSUserBundle at all and that the extension option is only for those who has a FOSUserBundle from before and wants to reuse some settings or such?

I could find no guidance on the issue of what the implications are of extending from a FOSUserBundle or not extending from a FOSUserBundle.

Th3Mouk:
In 2016 I have abandoned FosUserBundle in my projects.
All behaviors are simply realizable with the Symfony improvements in the security component.

By doing it ourselves we can drop some useless attributes in user class for example and reduce obscurity of usage. Simply get user features on our needs.
If we use it, its because this bundle was (is?) really popular and extendable, password recovery, emailing, rights and groups management already plugged with login form. And this behaviors already implemented, documented, are maintained outside of Sonata.

At that time it saved time and ensure robustness, was IMO a good choice, when I started with Sonata I was pretty confortable with this packed implementation.

In 2016, I personally think Sonata has (much) suffered from this bundle. More than the learning curve and the lack of doc of the version 2. Today it has become a brake on growth of the project and many new backend generator have emerged.
I don't want to carry the tick, but if porting FosUserBundle to Symfony4 is the same pain, Sonata will still suffer from this situation. 

I thank a lot, the maintainers who have evolved for years in the right direction this project.
If this site should start, it would take a lot of time, to leave as much flexibility in user management to developers.

This is only my vision, not a decision or an internal discussion.

greg0ire:
Thanks @Th3Mouk ! So you're saying this was simply done to let user choose if they want to continue using FOSUser or not? I guess it doesn't make the sonata bundle independent of the FOS bundle with such a small commit, but maybe it's simply a first step towards this direction?

Th3Mouk:
It's a way of thinking, I don't know if its the right way. If a user want to implement directly security component, it's probably already possible, somehow. The only part where I don't know at all how sonata is linked with Fos is the case of ACL.
The two things to think about are how to plug a from scratch user to connexion page and form, and redact a tutorial to how to create an user management view (with potentially groups).

ethernal:
@Th3Mouk Hi, could you please elaborate on how to drop FoSUserBundle from existing project - I'd love to do that but I'm not sure what I'll need to do exactly. I am still on SF 2.8 and can't switch to latest bundles because of FOSUB - I'd like to drop the overhead and I don't mind developing some parts by myself.

cassianotartari:
To use the latest sonata packages I've extend by myself FosUserBundle, I've created my own UserBundle now I can use:
```json
"friendsofsymfony/user-bundle": "~2.0@dev"
```
I've just took a look into SonataUserBundle code and FosUserBundle docs.

I'm using a complex permission system, kind of a ACL giving permissions for groups, implemented using bitwise logic. Almost like unix system.
```php
    const PERMISSION_NONE = 0;
    const PERMISSION_DELETE = 1;
    const PERMISSION_WRITE = 2;
    const PERMISSION_READ = 4;
    const PERMISSION_DOWNLOAD = 8;
```

Th3Mouk:
@ethernal It's a speculation, actually on my last projects, remove SonataUser = remove Sonata at all, because it's faster to rely direcly on symfony, than adapt all the thinking to Sonata component. My needs was to profit Symfony 3.2 features at release.

My idea is Sonata don't need SonataUser to work, so you can implement Symfony security component or with Guard like usual, create a login form with sonata css, and configure your firewall. The only missing piece is the user (and group if necessary) management, but it can be declared like any other CRUD entity.

To send email, or reset password and whatever, it's simply a route/controller/form or something else you love.

The delicate part of this work could be the missing of SonataUser interfaces, if there is one usefull.

OskarStark:
@Th3Mouk would be cool to have this solution as a separate bundle/component to use sonata with a login/user but not with FOSUser, maybe you could find some time 😄

Th3Mouk:
@OskarStark yes absolutely, i'm currently busy with angular, react, reactive programming, docker, rabbit... The part where I need reflexion is how to give to developper, enough abstraction, for example with a double or triple provider of user for the login form, manage user rights. IMO this component will be more a great tutorial than a real package of behaviors ready to plug.
I will start a POC this week.

OskarStark:
> I will start a POC this week.

excellent 🍺

jontorrado:
You could use this bundle which is written in DDD: https://github.com/BenGorUser/UserBundle. The owners will also contribute here, they won't leave the bundle untagged for years like FOS has done (cc @benatespina)

TerjeBr:
I still have not got any answers to my questions.

What are the implications of extending or not extending from the FOSUserBundle?
What are the differences?

Th3Mouk:
@TerjeBr I haven't tested it but seems SonataUser it's tightly coupled to Fos, and I don't know if it's possible to work without.

The implications are to reuse usefull provided features like, password recovering, signin/signup, (document) model extensions, mail sending. 

> At that time it saved time and ensure robustness, was IMO a good choice, when I started with Sonata I was pretty confortable with this packed implementation.

We talk about a minus 5 year old decision, this component has been tested and used by millions of project versions. So it's was easy to simply extend it for user management.

IMO Sonata need a complete tutorial, "How to implement user management from scratch in Sonata", that's it. And we can add too some links toward usefull bundle post in this thread.

greg0ire:
@Th3Mouk : I think you just showed us a key to getting good software : write documentation from the start. I'd even go as far as saying that in a perfect world, we should:

1. write docs about code that is not written yet
2. review the docs
3. write tests
4. finally write the code
5. review the code

TerjeBr:
@Th3Mouk My questions was about https://github.com/sonata-project/SonataUserBundle/tree/add_support_for_fos_user2

If you read the readme.md file there you will see that the bundle can be installed in 2 ways, with and without giving 'FOSUserBundle' as an argument to the SonataUserBundle constructor.

My question was what was the implication of installing that branch one way or the other. What is the difference if you say in the constructor that it should inherit from the FOSUserBundle, or if you do not?

greg0ire:
@TerjeBr : again I have no idea, but [this](https://symfony.com/doc/master/bundles/inheritance.html#main) + some analysis of the code might help you find an answer.

TerjeBr:
Not helpful @greg0ire

greg0ire:
Sorry I annoyed you with my help.

Th3Mouk:
@TerjeBr 

> SonataUser it's tightly coupled to Fos, and I don't know if it's possible to work without.

This 2 ways installing is very old, I presume it no longer work

TerjeBr:
@Th3Mouk How can this 2 ways of installing be very old, when the only branch (as far as I know) it is present on (tree/add_support_for_fos_user2) is a relatively new branch?

Do you at all know what you are talking about?

greg0ire:
> Do you at all know what you are talking about?

Do you?

```
commit 374b9397f28ab52de15208486a497e98b855e9c5
Author: Thomas Rabaix <thomas.rabaix@ekino.com>
Date:   Tue Jun 21 18:56:17 2011 +0200

    Make the bundle parent optional

diff --git a/README.md b/README.md
index 2b75556..a01eebd 100644
--- a/README.md
+++ b/README.md
@@ -1,4 +1,12 @@
 # Integrate the FOS/UserBundle into the Sonata Project
 
     - AdminBundle : add user and group management
-    - EasyExtends : allows to generate Application level model
\ No newline at end of file
+    - EasyExtends : allows to generate Application level model
+
+
+# Installation
+
+    you have 2 otpions to initialize the SonataUserBundle, you can select which bundle SonataUserBundle extends
+
+        - new Sonata\UserBundle\SonataUserBundle('FOSUserBundle') : the bundle will extends ``FOSUserBundle``
+        - new Sonata\UserBundle\SonataUserBundle() : the bundle will NOT extends ``FOSUserBundle``
diff --git a/SonataUserBundle.php b/SonataUserBundle.php
index d1ef644..ed84767 100644
--- a/SonataUserBundle.php
+++ b/SonataUserBundle.php
@@ -15,5 +15,15 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 class SonataUserBundle extends Bundle
 {
+    protected $parent;
 
+    public function __construct($parent = null)
+    {
+        $this->parent = $parent;
+    }
+
+    public function getParent()
+    {
+        return $this->parent;
+    }
 }
\ No newline at end of file
```

> Tue Jun 21 18:56:17 2011 +0200

It looks indeed weird, but maybe you could quit being a dick to every nice soul that tries to help you?

jordisala1991:
So... how can we help? I would like to use Sf3.x but this bundle is holding back because of FOSUserBundle.

Either we support FOS 2.0, or we remove it and implement our security without it. IMO I prefer the latest, reasons said by @Th3Mouk. Should we create some kind of poll to decide? (I agree that FOS is no longer needed with the current symfony security, but we still should provide an implementation of SonataUser, it avoids a lot of work every time you start a project)

I want to help with this bundle, but dunno where to start.

greg0ire:
If we are that far from being free from FOSUserBundle, then the logical course of action would be to start a new, better bundle. By better, I mean that the same mistakes are not done. So here is what I would suggest, feel free to add more conditions / steps :

1. The bundle should be fully and understandable from the start, otherwise we will get yelled at by people like @TerjeBr (not a big deal *per se*, but it expresses how frustrated and angry our userbase can become when they see that no one understands some pieces of the code anymore).
2. Don't do any of the mistakes listed in [this presentation](https://jolicode.github.io/fosuserbundle-conf/#/8)
    1. FOSUser is an implementation, not a framework, which is what makes it painful. Might be better to have a bundle that needs more configuration and does less, rather than one that takes decisions that you do not agree with, like for example, the username field, which is not really necessary when you have an email field.
    2. The database fields that you don't use and cannot be removed.
    3. Does not allow multiple user models (Frontend user vs Backend user).
3. Do it in TDD (meaning we write the tests from the start, and most of all, **before** writing the code itself).

I think to achieve that, you should start a project without SonataUserBundle, implement your own user management, and see what parts can be reused. When you identify a part, try to keep it small, and see the code from your project as a prototype: don't copy it, rewrite it from scratch in TDD. Same goes with SonataUserBundle.

> You could use this bundle which is written in DDD: https://github.com/BenGorUser/UserBundle. The owners will also contribute here, they won't leave the bundle untagged for years like FOS has done (cc @benatespina)

You could also use this bundle and report back instead, it does look quite good, apart from the kernel registration step, which is quite involving:

```php
public function registerBundles()
{
    $bundles = [
        // ...

        // Dependencies required by the bundle, keep the order.
        // First bridges and then the UserBundle

        // Bridges
        new BenGorUser\TwigBridgeBundle\TwigBridgeBundle(),
        new BenGorUser\SymfonyRoutingBridgeBundle\SymfonyRoutingBridgeBundle(),
        new BenGorUser\SymfonySecurityBridgeBundle\SymfonySecurityBridgeBundle(),
        new BenGorUser\SwiftMailerBridgeBundle\SwiftMailerBridgeBundle(),
        new BenGorUser\DoctrineORMBridgeBundle\DoctrineORMBridgeBundle(),
        new BenGorUser\SimpleBusBridgeBundle\SimpleBusBridgeBundle(),
        new BenGorUser\SimpleBusBridgeBundle\SimpleBusDoctrineORMBridgeBundle(),

        // User bundle
        new BenGorUser\UserBundle\BenGorUserBundle(),
        // ...
    ];
}
```

There would be nothing wrong IMO to depend on another bundle again, but if we do, we should:

- thoroughly review it;
- contribute to it.

benatespina:
Hi,

firstly thanks for consider BenGorUserBundle as a FOSUserBundle alternative, really appreciate it.

BenGorUser is a user management ecosystem, that the bundle is not more than a Symfony bridge,  as it could exist the one for Laravel. This packages are designed keeping in mind some DDD approaches like that the user always must be in a valid domain state, so you can view how does not exist public setters that corrupt the domain valid state.

Perhaps, to make our bundle compatible with SonataUserBundle we need to build these setters or something else in this direction (I mean in the anemic models direction), but it could be feasible.

About kernel registration, you're right, but as I said, this packages are designed keeping in mind the extremely decoupled approach and everything is changeable by other implementations. On the other hand, there is no problem for create SimpleBenGorUserBundle package that it can be a wrapper of BenGorUserBundle's typical configuration to simplify the installation steps. Also, it can be avoid this package and add this boilerplate implicitly inside SonataUserBundle, whatever you want.

A good and robust user management application has many considerations and it's not a short-term project so, I don't think that you have to consider the start of this custom develop.

I would love that BenGorUserBundle would be used as part of SonataUserBundle so, I'll help you whatever you need.

Thanks @jontorrado to recommend our bundle! ;)

I will mention my friend and co-author of BenGorUser @gorkalaucirica, so that he can also expose his vision.

TerjeBr:
Hi.

I have had a brief look trough the BenGorUserBundle. It seems to be very slanted towards the use case where users sign up themselves and users can invite other users. Is it flexible enough to be used in f.ex. an corporate  environment where users can only be added or edited by certain super users, and no one can sign up themself or be invited? If the bundle is felxible enough to be adjusted/configured to both cases then may be it can be something to think about.

benatespina:
Yes it is perfectly valid, we can disabled by configuration the signup use case and make our use case as explained in this chapter https://github.com/BenGorUser/UserBundle/blob/master/docs/extending_create_custom_use_case.md

jontorrado:
What about this? We should start ASAP, keeping this bundle for 2.8.* is hurting a lot.

core23:
@jontorrado Please read the comments... We're still waiting for a stable FOSUserBundle 2.x release

jontorrado:
I'm talking about changing to another user bundle which won't break the
stability as fos does. That bundle has been like that for a year.

andrewmy:
Heads up, FOSUserBundle 2.0 is released!

jordisala1991:
Closing this one, it is already on master

OskarStark:
Closed via #869

PR #750, opened 2016-05-28
Redirect route can be empty after registration

core23:
Rebase of #735

### Changelog

```markdown
### Fixed
- Fixed a potential empty route after calling `RegistrationFOSUser1Controller::registerAction`
```

### Subject
The route could be empty when no historic route was found in the user session.

### DIFF
--- a/Controller/RegistrationFOSUser1Controller.php
+++ b/Controller/RegistrationFOSUser1Controller.php
@@ -62,11 +62,7 @@ class RegistrationFOSUser1Controller extends Controller
                     $this->get('session')->remove('sonata_basket_delivery_redirect');
                     $url = $this->generateUrl($route);
                 } else {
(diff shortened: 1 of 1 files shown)


core23:

OskarStark:
LGTM 👍

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -65,7 +65,7 @@ public function registerAction()
                     $url = $this->get('session')->get('sonata_user_redirect_url');
```
greg0ire: How about 
```
$url = $this->get('session')->get(
    'sonata_user_redirect_url', 
    $this->generateUrl('sonata_user_profile_show')
)
``` instead then ? I don't think `$url` can end up being null inside the if, only inside the else…

core23: Fixed

PR #755, opened 2016-05-29
Fix bug of Doctrine MongoDB support

Dropaq:
ref: #440 

It's a fix of severe bug.

Support of mongoDB by Sonata-User-Bundle was completely broken, because Doctrine was unable to find FOS\UserBundle\Resources\config\doctrine\User.mongodb.xml while it iterates through inheritance.

So basically FOSUserBundle has FOS\Model which is extended twice by FOS\Entity and by FOS\Document.

At the same time SonataUserBundle implementing similar logic, but Sonata\UserBundle\Model\User is extending FOS\Entity.
It really doesn't make any sense. The only reason of existence of this Model is preventing from copy-pasting common fields between Entity and Document, but the price of it is too high.

Since support of PHP 5.3 is necessary and usage of traits is impossible, I just copied methods from Entity into Document.
Also I've fixed logic of UserManagerInterface and UserManagerProxy which caused similar problem with serializer. So basically missed declaration of service for mongoDB resulted in Fatal Error on project update.
By now all methods of newly updated UserManagerInterface for Document\UserManager is replaced with stubs throwing an Exception. I suppose serialization wouldn't work for mongo, but looks like everything else is up an running.


greg0ire:
Related : #745

Soullivaneuh:
@Dropaq Thanks for your PR!

I would like to give a try to #745 before considering this one.

Dropaq:
@SonataCI no thanks. PR was opened for 2 years, and I did like 3 rebases already.

Soullivaneuh:
@Dropaq First of all, Sonata**CI** is a bot. Is your PR is conflicted, it will say that.

If we don't merge your PR already it's because we are waiting other to be resolved like #745.

I'm sorry for the waiting. You might keep this PR open and wait a ping from a human to continue.

Regards.

PR #760, opened 2016-06-01
use range in composer.json instead of ~

OskarStark:
### Subject

The build was failing, due to a new composer lint check

### Example

https://travis-ci.org/sonata-project/SonataUserBundle/jobs/134360668

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "sonata-project/easy-extends-bundle": "^2.1",
         "sonata-project/google-authenticator": "^1.0",
         "friendsofsymfony/rest-bundle": "^1.1",
-        "jms/serializer-bundle": "~0.13 || ^1.0",
+        "jms/serializer-bundle": "^0.13 || ^1.0",
         "nelmio/api-doc-bundle": "^2.4"
     },
     "require-dev": {


greg0ire:
The commit message says you're using a new syntax, but the diff shows you also allow ~0.11 and ~0.12, with no explanation. Is it a mistake or not? Please make it clear.

OskarStark:
updated @greg0ire

greg0ire:
I don't still get why you suddenly allow 0.11 and 0.12. This is what you need to clarify.

OskarStark:
> I don't still get why you suddenly allow 0.11 and 0.12. This is what you need to clarify.

damn you are totally right here... i mixed things up with the other PR, will fix it in a minute

greg0ire:
Looks good @OskarStark :)

OskarStark:
ping @Soullivaneuh is this correct in this case?

Soullivaneuh:
> ping @Soullivaneuh is this correct in this case?

:+1:

greg0ire:
Thanks @OskarStark !

OskarStark:
patch or not @Soullivaneuh ?

Soullivaneuh:
@OskarStark Normaly `pedantic` but this label will be set this night only. So patch for the moment,

PR #767, opened 2016-06-08
Apply XML lint

SonataCI:


Resources/translations/SonataUserBundle.lt.xliff
```diff
@@ -1,6 +1,6 @@
-<?xml version="1.0"?>
-<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
-    <file source-language="lt" datatype="plaintext" original="" >
+<?xml version="1.0" encoding="UTF-8"?>
```
OskarStark: should be no problem, but this is not just linting, it adds this attribute

greg0ire: This is probably due to the encoding option that is being used, so, quite understandable.

Soullivaneuh: I did a manual check on an internal project with admin-bundle, this change nothing. :+1:

PR #791, opened 2016-09-23
Fix for the 'Bad Credentials' translation problem

### LINKED ISSUE #621: translation not working

### LINKED ISSUE #788: Bare exception messages displayed on login page
When an error occures during login, the exception message gets simply displayed above the login form. In my case a database error occures and something like

> An exception occurred while executing 'SELECT ...

is visible.

This behaviour has two mayor drawbacks:
* Sensitive and/or confusing error messages are leaked to the user
* Symfony's nice error pages in dev env are not displayed, so the error is harder to track down (EDIT: Just noticed that the error does not get logged, too)

ejkun:
Note: This is a rebuild of PR #716 

I am targetting this branch, because this the is branch I'm currently using in my projects, later I will check if the problem happens in other branches and if this solution works on then.

Closes #788 
Fixes   #621 

## Changelog

```markdown

### Changed
- Changed implementation of loginAction in SecurityFOSUser1Controller.php
- Changed implementation of loginAction in AdminSecurityController.php
- Changed how the error message is translated on login.html.twig
- Changed how the error message is translated on base_login.html.twig

### Removed
- Removed translation for 'Bad credentials' message in SonataUserBundle.de.xliff

### Fixed
- Fixed a possible security risk as noticed in this [line](https://github.com/sonata-project/SonataUserBundle/blob/88a962818dd6218379ff1439183a15647837bda0/Controller/AdminSecurityController.php#L40)
```

## Subject
As related in issue #621 there was a problem with the error message in the login page, that was not being translated. That was a error with FOSUserBundle, that apparently will be fixed in the 2.0 version, but since there is no prediction to when it will be ready. I made some changes by myself based on the dev-master of FOSUserBundle.

I changed both security controllers and the templates for the login page to look more like the latest version of FOSUserBundle. Now the error message is translated with the 'security' catalogue. Also took out the Bad Credentials entry in one of the translation files since there is no need for it.

### DIFF
--- a/Controller/SecurityFOSUser1Controller.php
+++ b/Controller/SecurityFOSUser1Controller.php
@@ -14,6 +14,8 @@ namespace Sonata\UserBundle\Controller;
 use FOS\UserBundle\Controller\SecurityController;
 use Sonata\UserBundle\Model\UserInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\SecurityContextInterface;
 
 /**
  * Class SecurityFOSUser1Controller.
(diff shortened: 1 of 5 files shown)


Resources/translations/SonataUserBundle.de.xliff
```diff
@@ -514,10 +514,6 @@
                 <source>Gender</source>
                 <target>Geschlecht</target>
             </trans-unit>
-            <trans-unit id="Bad credentials">
-                <source>Bad credentials</source>
-                <target>Benutzername oder Passwort ungültig</target>
-            </trans-unit>
```
greg0ire: Why remove this?

ejkun: This translation was only a quickfix to the problem, and it would not work perfectly.
Now the exception message is translated with the security catalogue, so is Symfony who is handling the translation.

ejkun:
I forgot to mention that this is a rebuild of PR #716. I will add this info to the PR description.

greg0ire:
Please review @sonata-project/contributors , this bug is quite annoying

Controller/SecurityFOSUser1Controller.php
```diff
@@ -37,6 +39,36 @@ public function loginAction()
...
-        return parent::loginAction();
+        /*
+            Implementation of parent::loginAction
+            Needed for fixing the "Bad Credentials" translation problem
+            The code is a mix of the 2.0.0 version
+            and the 1.3.6 version of FOSUserBundle
+        */
+
+        $request = $this->container->get('request');
+        /* @var $request \Symfony\Component\HttpFoundation\Request */
```
OskarStark: phpdoc should be above the var and start with `/**` instead of `/*`

Resources/translations/SonataUserBundle.de.xliff
OskarStark: do we need to remove this from all translation files?

ejkun: I don't think this translation exist in other translation files. At least I can't find it in other files.

Controller/SecurityFOSUser1Controller.php
ejkun: Ok, I will fix it.

ejkun: Done

greg0ire:
Thanks @ejkun !

PR #800, opened 2016-10-24
Wrong redirect route?

auipga:
I am targetting this branch, because…

Closes #
Fixes #
## Changelog

``` markdown
### Added
- Added some `Class::newMethod` to do great stuff

### Changed

### Deprecated

### Removed

### Fixed

### Security
```
## To do

- [ ] Update the tests
- [ ] Add an upgrade note
## Subject

Seems more valid to me.
If I'm wrong, can somebody explain this?

### DIFF
--- a/Controller/RegistrationFOSUser1Controller.php
+++ b/Controller/RegistrationFOSUser1Controller.php
@@ -56,10 +56,10 @@ class RegistrationFOSUser1Controller extends Controller
                 $url = $this->generateUrl('fos_user_registration_check_email');
             } else {
                 $authUser = true;
-                $route = $this->get('session')->get('sonata_basket_delivery_redirect');
+                $route = $this->get('session')->get('sonata_user_redirect_url');
 
                 if (null !== $route) {
-                    $this->get('session')->remove('sonata_basket_delivery_redirect');
+                    $this->get('session')->remove('sonata_user_redirect_url');
                     $url = $this->generateUrl($route);
                 } else {
                     $url = $this->get('session')->get('sonata_user_redirect_url', $this->generateUrl('sonata_user_profile_show'));


greg0ire:
Why didn't you fill in the PR template?

core23:
Please update the PR template, so we can merge this @auipga

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -56,10 +56,10 @@ public function registerAction()
                 $url = $this->generateUrl('fos_user_registration_check_email');
             } else {
                 $authUser = true;
-                $route = $this->get('session')->get('sonata_basket_delivery_redirect');
+                $route = $this->get('session')->get('sonata_user_redirect_url');
 
                 if (null !== $route) {
-                    $this->get('session')->remove('sonata_basket_delivery_redirect');
+                    $this->get('session')->remove('sonata_user_redirect_url');
```
greg0ire: It looks indeed weird… It was introduced in 4dbf95fc5a103eaa364abfd3ce3d805e5f7dcda8

greg0ire: @eko , can you please comment on this?

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -56,10 +56,10 @@ public function registerAction()
...
                     $url = $this->generateUrl($route);
                 } else {
                     $url = $this->get('session')->get('sonata_user_redirect_url', $this->generateUrl('sonata_user_profile_show'));
```
greg0ire: This is very very similar. @auipga , maybe we should only keep the else part, right?

greg0ire: Let's translate in pseudo-code : 

```
get the user redirect url in session

if it is not null
	remove it from the session
	generate an url from it (so it is in fact a route, not an url)
else
	try to get it again anyway, and default to the profile show route
```

Sound pretty stupid, right?  :stuck_out_tongue: 

Let 's just do 
```php
<?php
$url = $this->get('session')->get(
    'sonata_user_redirect_url',
     $this->generateUrl('sonata_user_profile_show')
);
// and maybe this too
$this->get('session')->remove('sonata_user_redirect_url');
```

Controller/RegistrationFOSUser1Controller.php
eko: Hi @greg0ire,

Indeed, this was for the `ecommerce` bundle suite in order to let user come back on its basket when he choose to register from the basket page (https://github.com/sonata-project/ecommerce/blob/7e778d57d1229df3f73b453404f9e7360b0ca923/src/BasketBundle/Controller/BasketController.php#L53).

greg0ire: Thanks for the feedback, do you agree with my proposal below?

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -56,10 +56,10 @@ public function registerAction()
                 $url = $this->generateUrl('fos_user_registration_check_email');
             } else {
                 $authUser = true;
-                $route = $this->get('session')->get('sonata_basket_delivery_redirect');
+                $route = $this->get('session')->get('sonata_user_redirect_url');
```
greg0ire: This is bad, you're storing a url in a route.

greg0ire:
I made [an alternative PR](https://github.com/sonata-project/SonataUserBundle/pull/806). What do you think?

PR #804, opened 2016-11-15
Define non-BC annotations dynamically

greg0ire:
I am targetting this branch, because this is a BC patch

See https://github.com/sonata-project/SonataUserBundle/commit/4b6e29540eab072e40535386050cef5a5c68e960#commitcomment-19809506

## Changelog

```markdown
### Fixed
- compatibility with FOSRestBundle 2.x was improved
```

## To do

- [x] Fix the unit tests
- [ ] Test on a real application @lemorragia, help?
- [ ] Decide whether to add `NEXT_MAJOR` instruction ( @sonata-project/contributors  are we dropping FOSRestBundle in the next major version ?)

## Subject

FOSRestBundle 1.x and 2.x do not use the same property name to specify
whether a query parameter is an array / map .
This has a drawback: this parameter will not appear in the
documentation, or be usable in the sandbox. But now both versions of
the FOSRestBundle should be usable.

### DIFF
--- a/Controller/Api/GroupController.php
+++ b/Controller/Api/GroupController.php
@@ -66,7 +66,6 @@ class GroupController
      *
      * @QueryParam(name="page", requirements="\d+", default="1", description="Page for groups list pagination (1-indexed)")
      * @QueryParam(name="count", requirements="\d+", default="10", description="Number of groups by page")
-     * @QueryParam(name="orderBy", array=true, requirements="ASC|DESC", nullable=true, strict=true, description="Query groups order by clause (key is field, value is direction")
      * @QueryParam(name="enabled", requirements="0|1", nullable=true, strict=true, description="Enabled/disabled groups only?")
      *
      * @View(serializerGroups="sonata_api_read", serializerEnableMaxDepthChecks=true)
@@ -77,6 +76,20 @@ class GroupController
      */
     public function getGroupsAction(ParamFetcherInterface $paramFetcher)
     {
+        $orderByQueryParam = new QueryParam();
+        $orderByQueryParam->name = 'orderBy';
+        $orderByQueryParam->requirements = 'ASC|DESC';
+        $orderByQueryParam->nullable = true;
+        $orderByQueryParam->strict = true;
+        $orderByQueryParam->description = 'Query groups order by clause (key is field, value is direction)';
+        if (property_exists($orderByQueryParam, 'map')) {
+            $orderByQueryParam->map = true;
+        } else {
+            $orderByQueryParam->array = true;
+        }
+
+        $paramFetcher->addParam($orderByQueryParam);
+
         $supportedFilters = array(
             'enabled' => '',
         );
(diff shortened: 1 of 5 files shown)


greg0ire:
@lemorragia, @GuilhemN, please review.

Controller/Api/GroupController.php
```diff
@@ -77,6 +76,20 @@ public function __construct(GroupManagerInterface $groupManager, FormFactoryInte
      */
     public function getGroupsAction(ParamFetcherInterface $paramFetcher)
     {
+        $orderByQueryParam = new QueryParam();
+        $orderByQueryParam->name = 'orderBy';
+        $orderByQueryParam->requirements = 'ASC|DESC';
+        $orderByQueryParam->nullable = true;
+        $orderByQueryParam->strict = true;
+        $orderByQueryParam->description = 'Query groups order by clause (key is field, value is direction)';
+        if (property_exists('FOS\RestBundle\Controller\Annotations\QueryParam', 'map')) {
```
GuilhemN: The FQCN can be replaced by `$orderByQueryParam`.

greg0ire: Indeed! Fixed.

greg0ire: > Isn't the following clearer ?

To me both versions are as clear as the other, but this one is more complex (one more variable name).

greg0ire:
The tests fail b/c `addParam` is not part of the interface… I'm going to try to change the typehint.

greg0ire:
Most tests are fixed, but the build for the lowest versions does not work, `addParam` does not seem to exist :/

greg0ire:
I'll just bump the minimum requirement to `1.5.0`. It's a minor upgrade, should be ok.

greg0ire:
Tests need to be fixed.

lemorragia:
I tested it on a working application. 
notes:

- "view:    { annotations: true }" must be enabled in sensio_framework_extra config
- there's another fix to be done, which i think is not documented in the upgrade guide of fos-rest, or at least i couldn't find it...and it's equally tricky: serializerGroups has become an array, apparently,
FROM serializerGroups="sonata_api_read" TO serializerGroups={"sonata_api_read"} 

not fixing this will throw a "Type error: Argument 1 passed to FOS\RestBundle\Context\Context::addGroups() must be of the type array, string given"

- sonata media suffers from the same problem
- the ordering with the fixes above, in my application(even if i don't use sonata-user extensively) is  working and sonata tests are passing

greg0ire:
Can you provide a PR based on this one @lemorragia ? Or at least some code?

lemorragia:
Yes. I can provide the code both for user-bundle and media-bundle. I forked media-bundle and i already pushed the code in my repo in a similar "bc_annotations" branch. After this one i'm gonna make a PR for that. In this case, could you tell me what is the correct way to contribute to this pull request?

greg0ire:
If you want to get my commits : 

```
git checkout -b greg0ire-bc_annotations 3.x
git pull https://github.com/greg0ire/SonataUserBundle.git bc_annotations
```

Then you can add your own code on top of that, push it to your own branch, and create your own PR (I will close this one when you do so).

greg0ire:
Closed in favor of #805

PR #806, opened 2016-11-19
Deprecate sonata basket delivery redirect

greg0ire:
I am targetting this branch, because this is BC (new deprecation)

Closes #800

## Changelog

```markdown
### Deprecated
- Relying on the `sonata_basket_delivery_redirect` is deprecated and won't be supported anymore
```

## Subject

There is a strange, undocumented mechanism that allows people to be redirected to an arbitrary route after registration, based on a very specific session variable named `sonata_basket_delivery_redirect`. This could be useful, but if it is, it should be implemented in a simpler way, for instance, reusing the `sonata_user_redirect_url` variable, and check if it is already set before overwriting it with the referer.

### DIFF
--- a/Controller/RegistrationFOSUser1Controller.php
+++ b/Controller/RegistrationFOSUser1Controller.php
@@ -13,7 +13,6 @@ namespace Sonata\UserBundle\Controller;
 
 use FOS\UserBundle\Model\UserInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
-use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
@@ -68,7 +67,7 @@ class RegistrationFOSUser1Controller extends Controller
 
             $this->setFlash('fos_user_success', 'registration.flash.user_created');
 
-            $response = new RedirectResponse($url);
+            $response = $this->redirect($url);
 
             if ($authUser) {
                 $this->authenticateUser($user, $response);
--- a/Controller/RegistrationFOSUser1Controller.php
+++ b/Controller/RegistrationFOSUser1Controller.php
@@ -58,6 +58,12 @@ class RegistrationFOSUser1Controller extends Controller
                 $route = $this->get('session')->get('sonata_basket_delivery_redirect');
 
                 if (null !== $route) {
+                    // NEXT_MAJOR: remove the if block
+                    @trigger_error(<<<'EOT'
+Setting a redirect url in the sonata_basket_delivery_redirect session variable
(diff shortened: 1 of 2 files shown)


greg0ire:
Please review @eko @auipga

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -59,16 +58,25 @@ public function registerAction()
                 $route = $this->get('session')->get('sonata_basket_delivery_redirect');
 
                 if (null !== $route) {
+                    // NEXT_MAJOR: remove the if block
+                    $deprecationMessage = <<<'EOT'
```
core23: OTV :D

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -59,16 +58,25 @@ public function registerAction()
...
+                        'sonata_user_redirect_url',
+                        $this->generateUrl('sonata_user_profile_show')
+                    );
                 }
             }
 
             $this->setFlash('fos_user_success', 'registration.flash.user_created');
 
-            $response = new RedirectResponse($url);
+            $response = $this->redirect($url);
```
core23: Can we use this with symfony 2.3?

Controller/RegistrationFOSUser1Controller.php
greg0ire: Héhé, I know… but I couldn't get vim to color this file normally without it :/ I'll try harder.

Controller/RegistrationFOSUser1Controller.php
greg0ire: Let me check, but I think this is quite old.

greg0ire: It's here since [almost forever](https://github.com/symfony/symfony/commit/9c07e46d91b30236d89a98f738b7b7f3e2e5dfd1)

Controller/RegistrationFOSUser1Controller.php
greg0ire: Not sure it's a vim problem after all. I don't know how to format it, let's ask styleCI.

greg0ire: StyleCI is fine with this :/

Controller/RegistrationFOSUser1Controller.php
```diff
@@ -59,16 +58,25 @@ public function registerAction()
                 $route = $this->get('session')->get('sonata_basket_delivery_redirect');
 
                 if (null !== $route) {
+                    // NEXT_MAJOR: remove the if block
+                    @trigger_error(<<<'EOT'
+Setting a redirect url in the sonata_basket_delivery_redirect session variable
+is deprecated since 3.x and will no longer result in a redirection to this url in 4.0.
+EOT
+                    , E_USER_DEPRECATED);
```
greg0ire: @core23 : I get a warning from syntastic about "Space found before comma in function call", but php-cs-fixer (and StyleCI) are fine with it, so let's go with that.

PR #812, opened 2016-12-15
remove SecurityContextInterface it was removed in symfony 3.0

Hanmac:
I am targetting this branch, because its an internal fix and is BC.

## Changelog

```markdown
### Changed
replace SecurityContextInterface with TokenStorageInterface and/or AuthorizationCheckerInterface

### Removed
removed use of SecurityContextInterface which was dropped in symfony 3.0
```

## Subject

removed SecurityContextInterface and replace it with TokenStorageInterface and/or AuthorizationCheckerInterface

it is part of the work to get sonata ready for symfony 3.*

### DIFF
--- a/Block/AccountBlockService.php
+++ b/Block/AccountBlockService.php
@@ -18,7 +18,7 @@ use Sonata\UserBundle\Model\UserInterface;
 use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
-use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
  * Class AccountBlockService.
@@ -31,22 +31,22 @@ use Symfony\Component\Security\Core\SecurityContextInterface;
 class AccountBlockService extends BaseBlockService
 {
     /**
-     * @var ProfileMenuBuilder
+     * @var TokenStorageInterface
      */
-    private $securityContext;
+    protected $tokenStorage;
 
     /**
      * Constructor.
      *
-     * @param string                   $name
-     * @param EngineInterface          $templating
-     * @param SecurityContextInterface $securityContext
+     * @param string                $name
+     * @param EngineInterface       $templating
+     * @param TokenStorageInterface $tokenStorage
      */
-    public function __construct($name, EngineInterface $templating, SecurityContextInterface $securityContext)
+    public function __construct($name, EngineInterface $templating, TokenStorageInterface $tokenStorage)
     {
         parent::__construct($name, $templating);
 
-        $this->securityContext = $securityContext;
+        $this->tokenStorage = $tokenStorage;
     }
 
     /**
@@ -55,8 +55,8 @@ class AccountBlockService extends BaseBlockService
     public function execute(BlockContextInterface $blockContext, Response $response = null)
     {
         $user = false;
-        if ($this->securityContext->getToken()) {
-            $user = $this->securityContext->getToken()->getUser();
+        if ($this->tokenStorage->getToken()) {
+            $user = $this->tokenStorage->getToken()->getUser();
         }
 
         if (!$user instanceof UserInterface) {
--- a/Block/AccountBlockService.php
+++ b/Block/AccountBlockService.php
@@ -18,6 +18,7 @@ use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\SecurityContextInterface;
 
 /**
  * Class AccountBlockService.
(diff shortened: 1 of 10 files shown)


Block/AccountBlockService.php
```diff
@@ -31,22 +30,22 @@
...
      *
-     * @param string                   $name
-     * @param EngineInterface          $templating
-     * @param SecurityContextInterface $securityContext
+     * @param string                $name
+     * @param EngineInterface       $templating
+     * @param TokenStorageInterface $tokenStorage
      */
-    public function __construct($name, EngineInterface $templating, SecurityContextInterface $securityContext)
+    public function __construct($name, EngineInterface $templating, TokenStorageInterface $tokenStorage)
```
greg0ire: Not sure it is that simple… this interface does not exist in old versions of symfony that we still support… [it was introduced in sf 2.6](https://github.com/symfony/symfony/commit/b96778718507e520596279954937f75c181efc17).

What you can do is remove the type hinting, and inject either the old thing and the new thing, depending on the version of the security component, and add `NEXT_MAJOR` markers explaining what to do when dropping compatibility with old versions of symfony

Resources/config/admin.xml
```diff
@@ -2,7 +2,8 @@
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="sonata.user.editable_role_builder" class="Sonata\UserBundle\Security\EditableRolesBuilder">
-            <argument type="service" id="security.context"/>
+            <argument type="service" id="security.token_storage"/>
+            <argument type="service" id="security.authorization_checker"/>
```
greg0ire: This should be done in a dynamic way (in php, inside the extension)

Hanmac:
@greg0ire ah i didn't noticed that Sonata does still support Symfony 2.3 

my current plan to fix it:
drop it from the constructor and add typehint-less set* functions i can call in the Extension

is that the right way to do that? Or have you an example for that?

greg0ire:
The dependencies are not optional so you shouldn't use setters IMO. Instead, I would use calls to `replaceArgument` or something like that.

greg0ire:
I found an example :) : https://github.com/sonata-project/SonataAdminBundle/blob/240ecbd02d453f9acaf81e00112ff5d8dae4459b/Security/Handler/AclSecurityHandler.php#L79

Hanmac:
@greg0ire yeah i am currently copying parts of the AdminBundle too.

i just wonder why it doesn't crash when it can't find the uses above.

greg0ire:
[Looks like it's just the way php works](https://3v4l.org/6hTMd)

Block/AccountBlockService.php
```diff
@@ -30,21 +31,27 @@
...
      */
     private $tokenStorage;
 
     /**
      * Constructor.
      *
-     * @param string                $name
-     * @param EngineInterface       $templating
-     * @param TokenStorageInterface $tokenStorage
+     * NEXT_MAJOR: Go back to signature class check when bumping requirements to SF 2.6+.
```
greg0ire: "signature class check" => "type hinting"

Block/AccountBlockService.php
```diff
@@ -30,21 +31,27 @@
...
+     * @param EngineInterface                                $templating
+     * @param TokenStorageInterface|SecurityContextInterface $tokenStorage
      */
-    public function __construct($name, EngineInterface $templating, TokenStorageInterface $tokenStorage)
+    public function __construct($name, EngineInterface $templating, $tokenStorage)
     {
         parent::__construct($name, $templating);
 
+        if (!$tokenStorage instanceof TokenStorageInterface && !$tokenStorage instanceof SecurityContextInterface) {
+            throw new \InvalidArgumentException('Argument 1 should be an instance of Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface or Symfony\Component\Security\Core\SecurityContextInterface');
```
greg0ire: Not telling you to split the string since this is still [debatable](https://github.com/sonata-project/dev-kit/issues/218), but I think you should add least but the message on its own line cc @sagikazarmark 

```php
throw new \InvalidArgumentException(
    'Argument 1 should yadda yadda…'
);
```

Tests/Security/EditableRolesBuilderTest.php
```diff
@@ -15,17 +15,39 @@
 
 class EditableRolesBuilderTest extends \PHPUnit_Framework_TestCase
 {
+
+    public function getTokenStorageMock()
+    {
+        // Set the SecurityContext for Symfony <2.6
+        // TODO: Remove conditional return when bumping requirements to SF 2.6+
```
greg0ire: "TODO" => "NEXT_MAJOR"

Tests/Security/EditableRolesBuilderTest.php
```diff
@@ -15,17 +15,39 @@
...
+            return $this->getMock('Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface');
+
+        }
+        return $this->getMock('Symfony\Component\Security\Core\SecurityContextInterface');
+    }
+
+    public function getAuthorizationCheckerMock()
+    {
+        // Set the SecurityContext for Symfony <2.6
+        // TODO: Remove conditional return when bumping requirements to SF 2.6+
```
greg0ire: Here too please

Block/AccountBlockService.php
sagikazarmark: If it breaks the char limit anyway then doesn't matter IMO

greg0ire: The char limit is arbitrary, people have wildly different screen widths. So it probably makes it better for someone (while not being enough for many people). It's a best effort if you will.

greg0ire:
@sonata-project/contributors , please give this a final review

greg0ire:
Thanks @Hanmac !

Hanmac:
@greg0ire: i did update my local source and noticed we missed something:
"use Symfony\Component\DependencyInjection\Reference;" need to be added to SonataUserExtension.

how it got pass the tests is beyond me oO

greg0ire:
> how it got pass the tests is beyond me oO

Maybe this part isn't covered at all :/

Can you provide a PR to fix this? Also, if you could add tests, it would great.

PR #814, opened 2016-12-16
Update SonataUserExtension.php

Hanmac:
I am targetting this branch, because it fixed a missing use in #812.

## Changelog

```markdown
### Fixed
- add missing use for DependencyInjection\Reference

```

## Subject

fixed add missing use for DependencyInjection\Reference


greg0ire:
If you want to write some tests one day, here is an example : https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Tests/DependencyInjection/SonataCoreExtensionTest.php

greg0ire:
Thanks @Hanmac ! I changed the commit message, the "update somefile.php" part was not really necessary… people are always changing files.

PR #828, opened 2017-01-24
Fix italian translations

gremo:
Add some missing italian translations and fix some keys bad/not translated.


OskarStark:
Thank you @gremo 👍

PR #841, opened 2017-03-03
Fixed empty route after registration

core23:

## Changelog
```
### Fixed
 - Fixed empty route after registration
```

## Subject

The current solution doesn't check for valid routes (non-empty / non-null) after a registration. E.g. the stored `sonata_user_redirect_url` value could contain an empty value, if the referer is empty.

### DIFF
--- a/Controller/RegistrationFOSUser1Controller.php
+++ b/Controller/RegistrationFOSUser1Controller.php
@@ -65,11 +65,12 @@ EOT
                     $this->get('session')->remove('sonata_basket_delivery_redirect');
                     $url = $this->generateUrl($route);
                 } else {
-                    $url = $this->get('session')->get(
-                        'sonata_user_redirect_url',
-                        $this->generateUrl('sonata_user_profile_show')
-                    );
+                    $url = $this->get('session')->get('sonata_user_redirect_url');
                 }
+            }
+
+            if (!$url) {
+                $url = $this->generateUrl('sonata_user_profile_show');
             }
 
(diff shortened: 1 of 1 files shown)


Controller/RegistrationFOSUser1Controller.php
```diff
@@ -64,13 +65,14 @@ public function registerAction()
                     $this->get('session')->remove('sonata_basket_delivery_redirect');
                     $url = $this->generateUrl($route);
                 } else {
-                    $url = $this->get('session')->get(
-                        'sonata_user_redirect_url',
-                        $this->generateUrl('sonata_user_profile_show')
```
jordisala1991: Not sure if I understand this change, AFAIK this second parameter purpose is to be used as a fallback when `get('sonata_user_redirect_url')` returns an empty value

http://api.symfony.com/2.8/Symfony/Component/HttpFoundation/Session/Session.html#method_get

core23: That's not right: 
> The default value **if not found**

There is no check, if the value is empty or null. That's why we need this fix.

jordisala1991: You are right!

Just for anyone that was thinking like me: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Session/Attribute/AttributeBag.php#L83

jordisala1991:
Thanks @core23 !

PR #842, opened 2017-03-03
Added missing throw in change password process

core23:

## Changelog

```markdown
### Fixed
 - Added missing throw in change password process
```

## Subject

The exception was created but not thrown to the request handler for post processing.

### DIFF
--- a/Controller/ChangePasswordFOSUser1Controller.php
+++ b/Controller/ChangePasswordFOSUser1Controller.php
@@ -33,7 +33,7 @@ class ChangePasswordFOSUser1Controller extends Controller
     {
         $user = $this->getUser();
         if (!is_object($user) || !$user instanceof UserInterface) {
-            throw $this->createAccessDeniedException('This user does not have access to this section.');
+            throw new AccessDeniedException('This user does not have access to this section.');
         }
 
         $form = $this->get('fos_user.change_password.form');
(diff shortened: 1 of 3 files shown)


Controller/ChangePasswordFOSUser1Controller.php
```diff
@@ -33,7 +33,7 @@ public function changePasswordAction()
     {
         $user = $this->getUser();
         if (!is_object($user) || !$user instanceof UserInterface) {
-            $this->createAccessDeniedException('This user does not have access to this section.');
+            throw $this->createAccessDeniedException('This user does not have access to this section.');
```
jordisala1991: This change is correct, but there is a problem... this method we are calling is not present on SF < 2.5:

http://symfony.com/blog/new-in-symfony-2-5-create-the-correct-denied-http-exception

IMHO we could simply replace this call by:

`throw new AccessDeniedException('This user does not have access to this section.')`

but there could be other ways like implement this method as a fallback for SF 2.3:

https://github.com/symfony/symfony/blob/2.5/src/Symfony/Bundle/FrameworkBundle/Controller/Controller.php#L166

Or do an `if` to check if the method exists.

greg0ire: I'd go with the simplest solution, which is not deprecated.

core23: We're calling this method any three different classes at the moment:
https://github.com/sonata-project/SonataUserBundle/search?utf8=%E2%9C%93&q=createAccessDeniedException

jordisala1991: Then I would just use `throw new AccessDeniedException('This user does not have access to this section.')` on all of them

core23: Fixed that

jordisala1991:
Thanks @core23

PR #822, opened 2016-12-31
DevKit updates for master branch

SonataCI:

PR #843, opened 2017-03-04
Bump CoreBundle to ^3.2

jordisala1991:
I am targetting this branch, because this is BC.

## Subject

This will fix PR (https://github.com/sonata-project/SonataUserBundle/pull/822).


core23:
Don't understand why does travis pass in the current `3.x` branch

jordisala1991:
Because we are using PHP 5.3 on the `--prefer-lowest`.

On master we use PHP 7.0 , that is why it is using PHPUnit ~5.4 and the error triggers.

core23:
Thanks @jordisala1991

PR #839, opened 2017-03-02
Remove deprecations about form factory on SF2.8

jordisala1991:
I am targetting this branch, because this is BC.

## Changelog

```markdown
### Removed
- Removed deprecations about form factory on SF2.8
```

## Subject

Removed deprecations about service definitions, form factory changes from SF2.3 to SF2.7/2.8. To be compatible with SF3.0 at some point we need this change.

Depends on: #838

PR #846, opened 2017-03-05
Fixing wrong factory definition

jordisala1991:

## Subject

I did a PR earlier which removed some deprecations (#839). Without this change, I get the following error on cache clear: invalid class name ('form.factory').

Setting factory as a Reference fixes this.

### DIFF
--- a/DependencyInjection/SonataUserExtension.php
+++ b/DependencyInjection/SonataUserExtension.php
@@ -88,12 +88,12 @@ class SonataUserExtension extends Extension
         $profileFormDefinition = $container->getDefinition('sonata.user.profile.form');
         $registrationFormDefinition = $container->getDefinition('sonata.user.registration.form');
         if (method_exists($profileFormDefinition, 'setFactory')) {
-            $profileFormDefinition->setFactory(array('form.factory', 'createNamed'));
-            $registrationFormDefinition->setFactory(array('form.factory', 'createNamed'));
+            $profileFormDefinition->setFactory(array(new Reference('form.factory'), 'createNamed'));
+            $registrationFormDefinition->setFactory(array(new Reference('form.factory'), 'createNamed'));
         } else {
-            $profileFormDefinition->setFactoryClass('form.factory');
+            $profileFormDefinition->setFactoryClass(new Reference('form.factory'));
             $profileFormDefinition->setFactoryMethod('createNamed');
-            $registrationFormDefinition->setFactoryClass('form.factory');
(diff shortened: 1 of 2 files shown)


DependencyInjection/SonataUserExtension.php
```diff
@@ -88,12 +88,12 @@ public function load(array $configs, ContainerBuilder $container)
         $profileFormDefinition = $container->getDefinition('sonata.user.profile.form');
         $registrationFormDefinition = $container->getDefinition('sonata.user.registration.form');
         if (method_exists($profileFormDefinition, 'setFactory')) {
-            $profileFormDefinition->setFactory(array('form.factory', 'createNamed'));
-            $registrationFormDefinition->setFactory(array('form.factory', 'createNamed'));
+            $profileFormDefinition->setFactory(array(new Reference('form.factory'), 'createNamed'));
+            $registrationFormDefinition->setFactory(array(new Reference('form.factory'), 'createNamed'));
         } else {
-            $profileFormDefinition->setFactoryClass('form.factory');
+            $profileFormDefinition->setFactoryClass(new Reference('form.factory'));
```
greg0ire: Why not use the fqcn here?

core23: Why should we use FQCNs here?

greg0ire: Because it's called "factory class". I didn't check the docs, but I think it makes sense, right?

jordisala1991: AFAIK both ways works, on MediaAdmin there is the other way: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/DependencyInjection/SonataMediaExtension.php#L49

On this bundle we used a reference on previous version: https://github.com/sonata-project/SonataUserBundle/blob/3.2.1/Resources/config/form.xml#L5

That's why I left it with a Reference... to preserve original code

If you want me to change it, I can do it.

greg0ire: I just noticed that this is code for old versions of sf… so nevermind, it does not need to be perfect b/c it will be removed soon (at least I hope)

jordisala1991: The other code we want maybe to change is this one: https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/config/form.xml#L7

greg0ire: Yes please do change this one.

jordisala1991:
Changed the new definition (that will be uncommented on next major).

Please review @core23 @greg0ire

PR #850, opened 2017-03-08
Master

torbec:
I am targeting this branch, because find a bug.

## Changelog

```markdown
### Added

### Changed

### Deprecated

### Removed

### Fixed
- SecurityFOSUser1Controller:Login
### Security
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

## Subject
wrong copy&past in Stop using deprecated csrf token generation (#845)

PR #845, opened 2017-03-05
Stop using deprecated csrf token generation

jordisala1991:
I am targetting this branch, because this is BC.

## Changelog

```markdown
### Fixed
- Avoid deprecation message by changing csrf token generation when possible
```

## To do

- [x] Update the tests

## Subject

To increase SF 3 compatibility, I changed the way we generate CSRF tokens with a BC layer.

Also I added some missing test classes, to start increasing code coverage, which is really low at the moment.

PR #851, opened 2017-03-08
Fix container calls when not extending from Controller

jordisala1991:
I am targeting this branch, because it is fixing a bug.

Closes https://github.com/sonata-project/SonataUserBundle/pull/850

## Changelog

```markdown
### Fixed
- Fix container calls when not extending from Controller
```

## Subject

This change was wrong: https://github.com/sonata-project/SonataUserBundle/pull/845/files#diff-b773cf909994196ea448ba8b717c1c55R72

We are not extending from Controller but from ContainerAware (FOSUser SecurityController extends from it) so we are not able to use `has` and `get` functions directly, we need to user `container` property.

### DIFF
--- a/Controller/SecurityFOSUser1Controller.php
+++ b/Controller/SecurityFOSUser1Controller.php
@@ -80,12 +80,10 @@ class SecurityFOSUser1Controller extends SecurityController
             ? Security::LAST_USERNAME : SecurityContextInterface::LAST_USERNAME;
 
         // NEXT_MAJOR: Symfony <2.4 BC. To be removed.
-        if ($this->has('security.csrf.token_manager')) {
-            $csrfToken = $this->get('security.csrf.token_manager')->getToken('authenticate')->getValue();
+        if ($this->container->has('security.csrf.token_manager')) {
+            $csrfToken = $this->container->get('security.csrf.token_manager')->getToken('authenticate')->getValue();
         } else {
-            $csrfToken = $this->has('form.csrf_provider')
-                ? $this->get('form.csrf_provider')->generateCsrfToken('authenticate')
-                : null;
+            $csrfToken = $this->container->get('form.csrf_provider')->generateCsrfToken('authenticate');
         }
 
         return $this->renderLogin(array(


Controller/SecurityFOSUser1Controller.php
```diff
@@ -80,12 +80,10 @@ public function loginAction()
...
         // NEXT_MAJOR: Symfony <2.4 BC. To be removed.
-        if ($this->has('security.csrf.token_manager')) {
-            $csrfToken = $this->get('security.csrf.token_manager')->getToken('authenticate')->getValue();
+        if ($this->container->has('security.csrf.token_manager')) {
+            $csrfToken = $this->container->get('security.csrf.token_manager')->getToken('authenticate')->getValue();
         } else {
-            $csrfToken = $this->has('form.csrf_provider')
-                ? $this->get('form.csrf_provider')->generateCsrfToken('authenticate')
-                : null;
+            $csrfToken = $this->container->get('form.csrf_provider')->generateCsrfToken('authenticate');
```
greg0ire: Why are you removing the ternary?

jordisala1991: This is the original code changed: https://github.com/sonata-project/SonataUserBundle/pull/845/files#diff-0f7ec6fc41c7f2a56dd9389f2841d511L81

greg0ire: OK… so?

jordisala1991: That ternary was copy pasted from the other code I did change.

Can bring it back, but I thought this would be more like the old code was.

greg0ire: Oh. Right. I get it know, keep it like this please!

OskarStark:
Same like here @jordisala1991 ?
https://github.com/sonata-project/SonataUserBundle/pull/852

jordisala1991:
Yes

greg0ire:
Why close this?

jordisala1991:
The other PR got merged first. I can keep it open to change the ternary

greg0ire:
Oh right

PR #852, opened 2017-03-09
Fix non-use of container for has/get services.

dapanas:
I am targeting this branch, because I'm adding a patch and is backward compatible. Not a major change.

## Changelog

```markdown
### Fixed
- Fix non-use of container for has/get services.
```

## Subject

In commit 728211902d19834554bed108f5c4a8ae3f70e337 an error was introduced in file *Controller/SecurityFOSUser1Controller.php* using `$this->has` and `$this->get` instead of `$this->container->has` and `$this->container->get`. This PR fixes that.


OskarStark:
Thank you @dapanas 👍

PR #855, opened 2017-03-21
Move sonata routes above fos

ZRazor:
I am targeting this branch, because its BC.

## Changelog

```markdown

### Changed
- Moved sonata routes above fos
```

## Subject

When fos routes are above sonata's, if we navigate for example to `sonata_user_profile_show`, it will be detected as `fos_user_profile_show`. It influences on KnpMenu. It does not detect `sonata_user_profile_show` as current url and does not highlight it.


jordisala1991:
@ZRazor do you rely on FOSUser1 Controllers on the SonataUserBundle (instead of using directly FOSUser controllers)? 

Is for any specific reason? I am planning on deprecating them. Thats why it would be nice to have your opinion on this

ZRazor:
I'm using `sonata_` routes, and they use FOSUser1 Controllers. There is no specific reason for it. If you  remove and deprecate them, I'll easily change routes to `fos_`, if behavior stays the same.

greg0ire:
What should we do with this PR then?

jordisala1991:
Not sure, @ZRazor can you try using fos_user routes and controllers?

What if someone is relying on routes being placed the other way around?

ZRazor:
@jordisala1991, I have already overridden everything I need, so this PR is not critical for me. I thought it was the right way to use `sonata_` routes, but apparently its not.

PR #857, opened 2017-03-23
Removed block bundle deprecations

core23:

## Changelog

```markdown
### Deprecated
 - Removed block service deprecation
```

## Subject

Removed sonata block bundle deprecation.

### DIFF
--- a/Block/AccountBlockService.php
+++ b/Block/AccountBlockService.php
@@ -11,8 +11,8 @@
 
 namespace Sonata\UserBundle\Block;
 
-use Sonata\BlockBundle\Block\BaseBlockService;
 use Sonata\BlockBundle\Block\BlockContextInterface;
+use Sonata\BlockBundle\Block\Service\AbstractBlockService;
 use Sonata\UserBundle\Model\UserInterface;
 use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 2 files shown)


Block/AccountBlockService.php
```diff
@@ -25,7 +25,7 @@
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-class AccountBlockService extends BaseBlockService
+class AccountBlockService extends AbstractBlockService
```
jordisala1991: Wait, shouldn't this be AbstractAdminBlockService?

composer.json
```diff
@@ -37,6 +37,7 @@
         "nelmio/api-doc-bundle": "^2.4",
         "sllh/php-cs-fixer-styleci-bridge": "^2.0",
         "sonata-project/seo-bundle": "^2.0",
+        "sonata-project/block-bundle": "^3.2",
```
jordisala1991: This needs to be before `seo-bundle`

Block/AccountBlockService.php
core23: No, because we only use normal block options. There ain't any `buildCreateForm` or `buildEditForm` method.

jordisala1991:
Thanks @core23

Soullivaneuh:
@core23 @OskarStark Deprecation reduction is a fix. Changelog and label are updated.

Block/AccountBlockService.php
greg0ire: @core23 I think @jordisala1991 was right, see https://github.com/sonata-project/SonataPageBundle/issues/841

core23: Maybe we should introduce a new interace for all embeddable blocks?

greg0ire: I'm not very familiar with the block bundle, what would make a block embeddable? The `getBlockMetadata` method ?

core23: Just figured out that there is already an interface for this `AdminBlockServiceInterface`

PR #859, opened 2017-03-23
Remove phpunit whitelist exceptions

jordisala1991:
I am targeting this branch, because this is pedantic.

## Subject

I think coverage is not calculated correctly, lets see

### DIFF
--- a/DependencyInjection/SonataUserExtension.php
+++ b/DependencyInjection/SonataUserExtension.php
@@ -84,7 +84,7 @@ class SonataUserExtension extends Extension
             $authorizationCheckerReference = new Reference('security.context');
         }
 
-        // NEXT_MAJOR: Remove following lines
+        // NEXT_MAJOR: Remove following lines.
         $profileFormDefinition = $container->getDefinition('sonata.user.profile.form');
         $registrationFormDefinition = $container->getDefinition('sonata.user.registration.form');
         if (method_exists($profileFormDefinition, 'setFactory')) {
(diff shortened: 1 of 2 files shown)


DependencyInjection/SonataUserExtension.php
```diff
@@ -84,7 +84,7 @@ public function load(array $configs, ContainerBuilder $container)
             $authorizationCheckerReference = new Reference('security.context');
         }
 
-        // NEXT_MAJOR: Remove following lines
+        // NEXT_MAJOR: Remove following lines.
```
jordisala1991: This is to trigger build

phpunit.xml.dist
```diff
@@ -12,23 +12,26 @@
          bootstrap="Tests/bootstrap.php"
 >
     <testsuites>
-        <testsuite name="User Bundle Test Suite">
-            <directory>./Tests/</directory>
+        <testsuite name="UserBundle Test Suite">
```
greg0ire: "UserBundle" isn't a thing, I think it was better before. "SonataUserBundle" is our user bundle.

jordisala1991: I will change all phpunit config files. Because none of them include word Sonata. I agree with you on the namings

jordisala1991: Changed to SonataUserBundle

jordisala1991:
Can we merge this?

greg0ire:
Sure

PR #861, opened 2017-03-25
Remove FOSUser controllers for FOSUser2

jordisala1991:
I am targeting this branch, because I want to remove them on master after upgrade to 2.0 (that's why first here) It will need a deprecation on 3.x.

## Subject

Having those controllers here only creates more maintenance issues... They are not required for the bundle to work.

Also having to define a Parent bundle it is not required


jordisala1991:
I removed those templates that are useles now because we have less controllers.

And fixed a bug when editing and showing user profile (id parameter was incorrect)

jordisala1991:

core23:
Do we still need this template? https://github.com/sonata-project/SonataUserBundle/blob/add_support_for_fos_user2/Resources/views/layout.html.twig

jordisala1991:
😱  not sure! I will have to test

PR #920, opened 2017-10-09
Drop old versions of php/sf

jlamur:

## Changelog

```markdown
### Removed
- Support for old versions of php and Symfony.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^5.3 || ^7.0",
+        "php": "^7.1",
         "friendsofsymfony/user-bundle": "^1.3",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/core-bundle": "^3.2",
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @jlamur !

composer.json
```diff
@@ -17,18 +17,18 @@
         }
     ],
     "require": {
-        "php": "^5.3 || ^7.0",
+        "php": "^7.1",
```
chernecov: "^2.8" didn't require 7.1, why to change version of php?
Even symfony 3.4 supports php 7.0....

greg0ire: I think we reverted that, didn't we?

chernecov: I didn't find revert commit.
Thing is that I can use 4.x version, because of php7.0 (I have restrictions for php upgrade)
Version 3.x I can use because of symfony 2.8 (I have symfony 3.4)

And as I see in 4.x : void types are already in use. :)

chernecov: Found "Restore php 5.6 through 7.0 compatibility", but symfony version is 2.8 there, so it doesn't suits me too.

jordisala1991: Yes, that happened to me aswell, you need to either use SF 2.8 and php 5.6  or upgrade php to 7.1 to use SF 3.4

greg0ire: Uuuuh right... too bad. If I were you I'd upgrade to 7.1, 7.0 is not supported anymore, apart for security fixes: http://php.net/supported-versions.php

chernecov: Ok. I see.
Then I will try to continue with custom solution:
https://github.com/chernecov/SonataUserBundle/tree/symfony3-php70-compatibility
Will wait for company to choose higher version of php. 
Thanks a lot for quick answers.

PR #921, opened 2017-10-10
Manual merge

jordisala1991:
Just approve, don't do the actual merge.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": "^7.1",
         "friendsofsymfony/user-bundle": "^2.0",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/core-bundle": "^3.2",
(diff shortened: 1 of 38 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": "^7.1",
         "friendsofsymfony/user-bundle": "^2.0",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/core-bundle": "^3.2",
```
core23: What about dropping support for symfony 3.1, as it is unsupported by symfony?

jordisala1991: Changed to SF ^3.1 (it was 3.0 before)

jordisala1991: I think I have to keep to 3.0 to be consistent with other bundles where that change is already applied. Not sure what to do, but seems weird to me to have some bundles on 3.0 others on 3.1 and others on 3.2. WDYT @core23 @greg0ire ?

greg0ire: We should support every actively maintained version since the last LTS

core23: So, please change this to `3.2^`, 'cause 3.1 is unsupported https://symfony.com/roadmap?version=3.1#checker

jordisala1991: done @greg0ire @core23

composer.json
```diff
@@ -17,18 +17,18 @@
...
         "sonata-project/easy-extends-bundle": "^2.2",
         "swiftmailer/swiftmailer": "^4.3 || ^5.0 || ^6.0",
-        "symfony/console": "^2.8 || ^3.0",
-        "symfony/form": "^2.8 || ^3.0",
-        "symfony/http-foundation": "^2.8 || ^3.0",
-        "symfony/security": "^2.8 || ^3.0"
+        "symfony/console": "^2.8 || ^3.2",
+        "symfony/form": "^2.8 || ^3.2",
+        "symfony/http-foundation": "^2.8 || ^3.2",
+        "symfony/security": "^2.8 || ^3.2"
```
jordisala1991: Here is the support for SF 2.8 @greg0ire

PR #926, opened 2017-10-16
Drop PHPUnit 4 support

vyshkant:
I am targeting this branch, because this PR does not break BC.

## Subject

Removed compatibility with PHPUnit 4. Fixes: sebastianbergmann/phpunit#2809


greg0ire:
Thanks @vyshkant !

PR #928, opened 2017-10-17
Manual merge

greg0ire:
cc @vyshkant


vyshkant:
@greg0ire I think you made a mistake while merging thees branches: I can't see the modification of `composer.json` I'd made (I'd changed the version of "matthiasnoback/symfony-dependency-injection-test" to "^2.2")

I think it could help.

greg0ire:
@vyshkant oh right! This does not play well with the other mnoback dependency sadly. Can you have a look?

greg0ire:
Here is how to get my code: 

```
git checkout -b greg0ire-master master
git pull https://github.com/greg0ire/SonataUserBundle.git master
```

vyshkant:
I think the problem is that the `matthiasnoback/SymfonyConfigTest` does not support PHPUnit 5.7.

https://github.com/matthiasnoback/SymfonyConfigTest/blob/e1539aa8552126b9ce41852ab23957a2315f1709/composer.json#L19-L21

And the `matthiasnoback/SymfonyDependencyInjectionTest` too.

https://github.com/matthiasnoback/SymfonyDependencyInjectionTest/blob/434ed11fd16b90b2432954c777d6162690a8048c/composer.json#L22-L24

So if to be consistent, we must either refuse to use the `matthiasnoback/*`, or switch to a PHPUnit 6.

vyshkant:
Or we can continue to use older versions, in which case I need to rewrite the tests.

greg0ire:
I'd rather you use that last solution, switching to phpunit 6 is not something we can do at the very short term I think.

vyshkant:
@greg0ire Where should I commit my changes?

greg0ire:
on the stable branch, let's keep the difference minimal please.

vyshkant:
> on the stable branch, let's keep the difference minimal please.

See #929

greg0ire:
Wow @core23 you're an early bird!

PR #930, opened 2017-10-18
Configuration default values moved to the Configuration class

### LINKED ISSUE #927: Move configuration defaults to the Configuration class
### Subject

`Sonata\UserBundle\DependencyInjection\SonataUserExtension` class contains the following method, which is called from `Sonata\UserBundle\DependencyInjection\SonataUserExtension::load` method.

https://github.com/sonata-project/SonataUserBundle/blob/258192330405d49b5fc42be090572f6b3c5ac14a/DependencyInjection/SonataUserExtension.php#L208-L230

Please pay special attention to the following lines

https://github.com/sonata-project/SonataUserBundle/blob/258192330405d49b5fc42be090572f6b3c5ac14a/DependencyInjection/SonataUserExtension.php#L223-L224

Here I see several problems at once:
1. We have hardcoded class names, and we have **absolutely no guarantees** that these classes exist.
2. We use these class names as default values, without forcing the user to enter the actual values, while these default hardcoded values themselves may be incorrect.
3. We do all this in class `Sonata\UserBundle\DependencyInjection\SonataUserExtension`, so the user (and the `Sonata\UserBundle\DependencyInjection\Configuration` class) has no idea that these properties have default values.

### An example of a problem that I encountered

I'm not using `Application\Sonata\UserBundle...` class path for overridden user bundle in my application. Instead, I use `Overridden\Sonata\UserBundle...`, so the `Application` keyword does not concern me at all (**and is incorrect!**), but the code shown above makes it default. The `Configuration` does not force me to fill this values in config.

### What I suggest

1. Put the default configuration setting into the class `Configuration`.
2. Make default values to be **valid**, for example, using the standard, not overridden Entities (`Sonata\UserBundle\Entity\BaseUser` etc).
3. Make `SonataUserExtension::addDefaults` empty and remove in next major.
4. Modify the `SonataUserExtension::load` to check whether the `manager_type` config value corresponds to the `user` and `group` values.

vyshkant:
I am targeting this branch, because this PR breaks BC

Closes #927

## Changelog

```markdown
### Changed
- default values moved to the Configuration class
```

## To do

- [x] Update the tests

## Subject

Configuration default values moved from `SonataUserExtension::load` to the `Configuration` class.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -11,6 +11,10 @@
 
 namespace Sonata\UserBundle\DependencyInjection;
 
+use Sonata\UserBundle\Admin\Entity\GroupAdmin;
+use Sonata\UserBundle\Admin\Entity\UserAdmin;
+use Sonata\UserBundle\Entity\BaseGroup;
+use Sonata\UserBundle\Entity\BaseUser;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 
@@ -65,9 +69,10 @@ class Configuration implements ConfigurationInterface
                     ->end()
                 ->end()
                 ->arrayNode('class')
+                    ->addDefaultsIfNotSet()
                     ->children()
-                        ->scalarNode('group')->cannotBeEmpty()->end()
-                        ->scalarNode('user')->cannotBeEmpty()->end()
+                        ->scalarNode('group')->cannotBeEmpty()->defaultValue(BaseGroup::class)->end()
+                        ->scalarNode('user')->cannotBeEmpty()->defaultValue(BaseUser::class)->end()
                     ->end()
                 ->end()
                 ->arrayNode('admin')
@@ -76,7 +81,7 @@ class Configuration implements ConfigurationInterface
                         ->arrayNode('group')
                             ->addDefaultsIfNotSet()
                             ->children()
-                                ->scalarNode('class')->cannotBeEmpty()->end()
+                                ->scalarNode('class')->cannotBeEmpty()->defaultValue(GroupAdmin::class)->end()
                                 ->scalarNode('controller')->cannotBeEmpty()->defaultValue('SonataAdminBundle:CRUD')->end()
                                 ->scalarNode('translation')->cannotBeEmpty()->defaultValue('SonataUserBundle')->end()
                             ->end()
@@ -84,7 +89,7 @@ class Configuration implements ConfigurationInterface
                         ->arrayNode('user')
                             ->addDefaultsIfNotSet()
                             ->children()
-                                ->scalarNode('class')->cannotBeEmpty()->end()
+                                ->scalarNode('class')->cannotBeEmpty()->defaultValue(UserAdmin::class)->end()
                                 ->scalarNode('controller')->cannotBeEmpty()->defaultValue('SonataAdminBundle:CRUD')->end()
                                 ->scalarNode('translation')->cannotBeEmpty()->defaultValue('SonataUserBundle')->end()
                             ->end()
--- a/DependencyInjection/SonataUserExtension.php
+++ b/DependencyInjection/SonataUserExtension.php
@@ -12,6 +12,14 @@
 namespace Sonata\UserBundle\DependencyInjection;
 
 use Sonata\EasyExtendsBundle\Mapper\DoctrineCollector;
+use Sonata\UserBundle\Admin\Document\GroupAdmin as DocumentGroupAdmin;
+use Sonata\UserBundle\Admin\Document\UserAdmin as DocumentUserAdmin;
+use Sonata\UserBundle\Admin\Entity\GroupAdmin as EntityGroupAdmin;
+use Sonata\UserBundle\Admin\Entity\UserAdmin as EntityUserAdmin;
+use Sonata\UserBundle\Document\BaseGroup as DocumentGroup;
+use Sonata\UserBundle\Document\BaseUser as DocumentUser;
+use Sonata\UserBundle\Entity\BaseGroup as EntityGroup;
+use Sonata\UserBundle\Entity\BaseUser as EntityUser;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -81,7 +89,7 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
             $loader->load('security_acl.xml');
         }
 
-        $config = $this->addDefaults($config);
+        $this->checkManagerTypeToModelTypeMapping($config);
 
         $this->registerDoctrineMapping($config);
         $this->configureAdminClass($config, $container);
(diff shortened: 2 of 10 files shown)


DependencyInjection/Configuration.php
```diff
@@ -65,9 +69,10 @@ public function getConfigTreeBuilder()
                     ->end()
                 ->end()
                 ->arrayNode('class')
+                    ->addDefaultsIfNotSet()
                     ->children()
-                        ->scalarNode('group')->cannotBeEmpty()->end()
-                        ->scalarNode('user')->cannotBeEmpty()->end()
+                        ->scalarNode('group')->cannotBeEmpty()->defaultValue(BaseGroup::class)->end()
```
core23: This will work with the default entities?

vyshkant: > This will work with the default entities?

I see no reason why it wouldn't work.

core23: You're right. Thought this was an abstract class

vyshkant:
@sonata-project/contributors what do you think about this?

greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

vyshkant:
@greg0ire I'have updated the changelog.

greg0ire:
It's still very long: try to explain your PR in one sentence to someone very dumb.

DependencyInjection/SonataUserExtension.php
```diff
@@ -273,4 +257,49 @@ protected function aliasManagers(ContainerBuilder $container, $managerType)
         $container->setAlias('sonata.user.user_manager', sprintf('sonata.user.%s.user_manager', $managerType));
         $container->setAlias('sonata.user.group_manager', sprintf('sonata.user.%s.group_manager', $managerType));
     }
+
+    /**
+     * @param array $config
+     */
+    private function checkManagerTypeToModelTypeCorrespondence(array $config)
```
greg0ire: Please replace "Correspondence" with "Mapping"

DependencyInjection/SonataUserExtension.php
```diff
@@ -273,4 +257,49 @@ protected function aliasManagers(ContainerBuilder $container, $managerType)
...
+
+        foreach ($actualModelClasses as $index => $actualModelClass) {
+            if ('\\' === substr($actualModelClass, 0, 1)) {
+                $actualModelClass = substr($actualModelClass, 1);
+            }
+
+            $expectedModelClass = $expectedModelClasses[$index];
+
+            if ($actualModelClass !== $expectedModelClass && !is_subclass_of($actualModelClass, $expectedModelClass)) {
+                throw new \InvalidArgumentException(sprintf('Model class "%s" does not correspond to manager type "%s".', $actualModelClass, $managerType));
```
greg0ire: Please linebreak this.

vyshkant:
updated changelog

DependencyInjection/SonataUserExtension.php
```diff
@@ -273,4 +257,55 @@ protected function aliasManagers(ContainerBuilder $container, $managerType)
...
+            ];
+        } elseif ('mongodb' === $managerType) {
+            $expectedModelClasses = [
+                DocumentUser::class,
+                DocumentGroup::class,
+                DocumentUserAdmin::class,
+                DocumentGroupAdmin::class,
+            ];
+        } else {
+            throw new \InvalidArgumentException(sprintf('Invalid manager type "%s".', $managerType));
```
Soullivaneuh: It would be great to drop `sprintf` usage for string encapsulation instead. This is about performance.

- https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html
- https://jonczyk.me/2016/10/23/phpyths-buster-great-string-performance-test/

Maybe the rules should be updated on dev-kit.

DependencyInjection/SonataUserExtension.php
```diff
@@ -273,4 +257,55 @@ protected function aliasManagers(ContainerBuilder $container, $managerType)
...
+        }
+
+        foreach ($actualModelClasses as $index => $actualModelClass) {
+            if ('\\' === substr($actualModelClass, 0, 1)) {
+                $actualModelClass = substr($actualModelClass, 1);
+            }
+
+            $expectedModelClass = $expectedModelClasses[$index];
+
+            if ($actualModelClass !== $expectedModelClass && !is_subclass_of($actualModelClass, $expectedModelClass)) {
```
Soullivaneuh: This line needs a break. And maybe a KitKat.

DependencyInjection/SonataUserExtension.php
greg0ire: Personally, that's typically the type of string where I'd stick with `sprintf`. It's not about performance, it's about legibility, especially when you have double quotes like here.

DependencyInjection/SonataUserExtension.php
greg0ire: ba dum tss

DependencyInjection/SonataUserExtension.php
OskarStark: and so you really need performance when you misconfigure sth in dev mode?

I would like to stick with the sprintf()

OskarStark:
@vyshkant I updated your changelog

DependencyInjection/SonataUserExtension.php
Soullivaneuh: Well, let's go for it.

DependencyInjection/SonataUserExtension.php
vyshkant: @Soullivaneuh But its length is 119 characters. As far as I understand, this is acceptable, is not it? Maybe you mean the next line?
```php
                throw new \InvalidArgumentException(sprintf('Model class "%s" does not correspond to manager type "%s".', $actualModelClass, $managerType));
```

Soullivaneuh: Oh, indeed. Even if a line-break would be great here, I'll approve the PR so.

greg0ire:
Thanks @vyshkant !

jordisala1991:
Just to clarify because of the problem we found on this issue: https://github.com/sonata-project/SonataUserBundle/issues/942

The intention after this PR is to do the configuration always on symfony config like this?:

```
sonata_user:
    class:
        user: Application\Sonata\UserBundle\Entity\User
        group: Application\Sonata\UserBundle\Entity\Group
```

greg0ire:
Might be related to https://github.com/sonata-project/SonataUserBundle/issues/944 ?

jordisala1991:
yes it is related.. let's wait a little bit, but I am not sure if we should revert this PR or not

vyshkant:
The idea of this PR is that the values of `Application\Sonata\UserBundle\Entity\User` and `Application\Sonata\UserBundle\Entity\Group` are no longer default (because these classes may not even exist).

After this PR, their base classes — `Sonata\UserBundle\Entity\BaseUser` and `Sonata\UserBundle\Entity\BaseGroup`, which obviously exist) become default ones.

To use the child classes (in the example above — `Application\Sonata\UserBundle\Entity\User` and `Application\Sonata\UserBundle\Entity\Group`), you need to specify them in the config.

But I still don't understand what happened here: https://github.com/sonata-project/SonataUserBundle/issues/942.

Perhaps the solution of this problem is that we need to output another error message. Give me a little time to research the problem.

vyshkant:
Most likely, the problem is that the base classes are not suitable for use as default values, because they do not have an `id` property (as well as in their mapping it is also not present).

In this situation, I see two possible options:
1) completely remove the default values for these properties and force the user to specify overridden values;
2) add `id` property to the base classes (and to their mappings) to make base classes suitable for use as default values.

WDYT @greg0ire @jordisala1991 @Soullivaneuh

OskarStark:
@sonata-project/contributors please vote

jordisala1991:
Why the default value is not a string: ‘Application\Sonata\UserBundle...’?

Those classes exists if you install the bundle like we say in docs. If you do it different, then it makes sense to redefine that parameter.

What I mean is: we did 2 things on this PR:

1. Move config out of compiler pass to config.php which is great.
2. Change a default value that was working for most cases to a default value that will never work.

IMO the second part should be reverted. If not I do agree on having to define always those params, but I think it is an unnecesary configurarion.

vyshkant:
> Change a default value that was working for most cases to a default value that will never work.

And for the other part of cases, it will cause an error, I personally spent several hours searching for it.

I think that kind of class `Application\Sonata\UserBundle...` are absolutely arbitrary, and their presence in the code (not in the config) is the hardcode. According to my ideas, this must be avoided in every possible way.

greg0ire:
I voted

vyshkant:
I'm going to create a PR solving this https://github.com/sonata-project/SonataUserBundle/pull/930#issuecomment-342327896 problem.

Most likely this PR will violate BC, but at the same time it will fix the bug (the current default values are invalid). Invalid (error-causing) values will be removed and the user will be forced to specify valid values.

According to https://github.com/sonata-project/SonataUserBundle/blob/4.x/CONTRIBUTING.md, I have to choose `master` branch, but I have doubts: "BC-break" effect will only occur if invalid default values are used, and "BC-break" effect will not occur for a working configuration.

My question is: can I target `4.x` branch?

greg0ire:
I think that is acceptable.

vyshkant:
See https://github.com/sonata-project/SonataUserBundle/pull/981

PR #947, opened 2017-11-07
Logo was only shown in login template

OskarStark:
I am targeting this branch, because it is not a BC break!

## Changelog

```markdown
### Fixed
- missing logo in templates
```

### DIFF
--- a/Resources/views/Admin/Security/Resetting/checkEmail.html.twig
+++ b/Resources/views/Admin/Security/Resetting/checkEmail.html.twig
@@ -27,6 +27,11 @@ file that was distributed with this source code.
     <div class="login-box">
         <div class="login-logo">
             <a href="{{ path('sonata_admin_dashboard') }}">
+                {% if 'single_image' == adminPool.getOption('title_mode') or 'both' == adminPool.getOption('title_mode') %}
+                    <div>
+                        <img style="width:64px;" src="{{ asset(adminPool.titlelogo) }}" alt="{{ adminPool.title }}">
+                    </div>
+                {% endif %}
                 {% if 'single_text' == admin_pool.getOption('title_mode') or 'both' == admin_pool.getOption('title_mode') %}
                     <span>{{ admin_pool.title }}</span>
                 {% endif %}
--- a/Resources/views/Admin/Security/Resetting/checkEmail.html.twig
+++ b/Resources/views/Admin/Security/Resetting/checkEmail.html.twig
@@ -25,18 +25,20 @@ file that was distributed with this source code.
 {% block sonata_wrapper %}
 
     <div class="login-box">
-        <div class="login-logo">
-            <a href="{{ path('sonata_admin_dashboard') }}">
-                {% if adminPool.getOption('title_mode') in ['single_image', 'both'] %}
-                    <div>
-                        <img style="width:64px;" src="{{ asset(adminPool.titlelogo) }}" alt="{{ adminPool.title }}">
-                    </div>
-                {% endif %}
-                {% if adminPool.getOption('title_mode') in ['single_text', 'both'] %}
-                    <span>{{ admin_pool.title }}</span>
-                {% endif %}
-            </a>
-        </div>
+        {% block login_box_header %}
+            <div class="login-logo">
+                <a href="{{ path('sonata_admin_dashboard') }}">
+                    {% if adminPool.getOption('title_mode') in ['single_image', 'both'] %}
+                        <div>
+                            <img style="width:64px;" src="{{ asset(adminPool.titlelogo) }}" alt="{{ adminPool.title }}">
+                        </div>
+                    {% endif %}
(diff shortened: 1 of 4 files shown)


OskarStark:
StyleCi not related to this PR

Resources/views/Admin/Security/Resetting/checkEmail.html.twig
```diff
@@ -27,6 +27,11 @@ file that was distributed with this source code.
     <div class="login-box">
         <div class="login-logo">
             <a href="{{ path('sonata_admin_dashboard') }}">
+                {% if 'single_image' == adminPool.getOption('title_mode') or 'both' == adminPool.getOption('title_mode') %}
```
greg0ire: `if adminPool.getOption('title_mode') in ['single_image', 'both']`

Resources/views/Admin/Security/Resetting/checkEmail.html.twig
```diff
@@ -27,6 +27,11 @@ file that was distributed with this source code.
...
+                    <div>
+                        <img style="width:64px;" src="{{ asset(adminPool.titlelogo) }}" alt="{{ adminPool.title }}">
```
greg0ire: Uuuuh no way around the `style` attribute? Like a class?

jlamur:
For reference, on the login page it's done this way:
https://github.com/sonata-project/SonataUserBundle/blob/c05b4ff321f65b5c34194bda1ea0955e099a8525/src/Resources/views/Admin/Security/login.html.twig#L29-L37

jlamur:
If it's possible, please create a new block to reuse it in all templates that didn't have the logo.

You forgot to add the logo to these pages:
- views/Admin/Security/Resetting/passwordAlreadyRequested.html.twig,
- views/Admin/Security/two_step_form.html.twig (this one seems to be completly different, so I'm not sure if it should be modified or not).

Resources/views/Admin/Security/Resetting/checkEmail.html.twig
jlamur: Maybe it's time create a `title_logo_size` option?
I'm not proud but I've overidden the whole template because I needed to have a wider logo :)

greg0ire: Do we really have to use CSS resize, BTW?

greg0ire:
> For reference, on the login page it's done this way

Just because the code is old doesn't mean it is good :P

Resources/views/Admin/Security/Resetting/checkEmail.html.twig
jlamur: If there is a way to delete this `width` and that users who have a logo wider than 64px don't see it at their original size, then please remove it :)

greg0ire: `overflow: hidden`?

core23:
> I am targeting this branch, because is not BC

Why?

OskarStark:
> Why?

@core23 , what do you mean? the templates in 3.x look different and from a technical perspektive its not a BC break. Ahhh you man it **is** BC, right? 😄

OskarStark:
> views/Admin/Security/Resetting/passwordAlreadyRequested.html.twig,

this one isn't available in `master` branch

> views/Admin/Security/two_step_form.html.twig (this one seems to be completly different, so I'm not sure if it should be modified or not).

@jlamur you are right, this one looks totally different, and therefore I ignored it...

Resources/views/Admin/Security/Resetting/checkEmail.html.twig
OskarStark: done

OskarStark:
> If it's possible, please create a new block to reuse it in all templates that didn't have the logo.

which block name do you prefer?

core23:
> what do you mean? the templates in 3.x look different and from a technical perspektive its not a BC break. Ahhh you man it **is** BC, right? 😄 

That's what I mean. There is no break, it's a new feature. So you could target the stable branch for this.

OskarStark:
> That's what I mean. There is no break, it's a new feature. So you could target the stable branch for this.

indeed, I checked it, but the templates look completely different, so i decided, that it make not much sense. Can you confirm this @jordisala1991 ?

> @OskarStark Maybe login_box_logo?

yes but we should wrap the whole "header" with logo **and** title instead, shouldn't we?

jordisala1991:
There are less templates on 3.x, but the ones in master that are there on 3.x should be quite similar @OskarStark

OskarStark:
This is the login template on `3.x` and looks totally different:
https://github.com/sonata-project/SonataUserBundle/blob/3.x/src/Resources/views/Security/base_login.html.twig

OskarStark:
are you guys 👍  for the new block?

OskarStark:
I added `login_box_header` block

greg0ire:
Thanks @Oskarstark

PR #948, opened 2017-11-07
Indention

OskarStark:
I am targeting this branch, because this is a doc fix.


jlamur:
Isn't it called *indentation* in english? https://en.wikipedia.org/wiki/Indentation_(typesetting)

Soullivaneuh:
@jlamur You are correct. The commit message may be fixed while merging.

greg0ire:
@jlamur @Soullivaneuh Both work

greg0ire:
And I usually use "indention", I believe it is more widely used.

PR #950, opened 2017-11-09
Changed SonataUserExtension consistency check

### LINKED ISSUE #944: Model class "Application\Sonata\UserBundle\Admin\UserAdmin" does not correspond to manager type "orm".
```yaml
sonata_user:
    security_acl: true
    manager_type: orm # can be orm or
    class:
        user: Application\Sonata\UserBundle\Entity\User
        group: Application\Sonata\UserBundle\Entity\Group
    admin:                  # Admin Classes
        user:
            class:          Application\Sonata\UserBundle\Admin\UserAdmin # there define your extended admin class
            controller:     SonataUserBundle:UserCURD
            translation:    SonataUserBundle
```
get this above error while extending user Admin class to change in user form

vyshkant:
I am targeting this branch, because this is a fix of created in this branch method.

Closes #944

## Changelog

```markdown
### Fixed
- Fixed bug with inability to use custom admin class
```

## Subject

Existing consistency checks (`SonataUserExtension::checkManagerTypeToModelTypeMapping`) led to the inability to use custom (not inherited from our base) admin class.

### DIFF
--- a/DependencyInjection/SonataUserExtension.php
+++ b/DependencyInjection/SonataUserExtension.php
@@ -12,10 +12,6 @@
 namespace Sonata\UserBundle\DependencyInjection;
 
 use Sonata\EasyExtendsBundle\Mapper\DoctrineCollector;
-use Sonata\UserBundle\Admin\Document\GroupAdmin as DocumentGroupAdmin;
-use Sonata\UserBundle\Admin\Document\UserAdmin as DocumentUserAdmin;
-use Sonata\UserBundle\Admin\Entity\GroupAdmin as EntityGroupAdmin;
-use Sonata\UserBundle\Admin\Entity\UserAdmin as EntityUserAdmin;
 use Sonata\UserBundle\Document\BaseGroup as DocumentGroup;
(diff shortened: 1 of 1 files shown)


DependencyInjection/SonataUserExtension.php
```diff
@@ -307,5 +316,23 @@ private function checkManagerTypeToModelTypeMapping(array $config)
...
+                    throw new \InvalidArgumentException(
+                        sprintf(
+                            'Admin class "%s" does not correspond to manager type "%s".',
+                            $actualAdminClass,
+                            $managerType
+                        )
+                    );
+                }
+            }
+        }
```
greg0ire: I don't think that check is very useful, and it doesn't scale well.

vyshkant: Actually yes. It is better to remove it?

greg0ire: Yeah. Or maybe introduce marker interfaces and trigger a deprecation if they are not implemented, but meh

vyshkant: Ok. I'm removing :)

PR #954, opened 2017-11-18
Prepare for symfony 3.4

Jan0707:


.travis.yml
```diff
@@ -69,6 +69,8 @@ matrix:
     - env: FOS_USER=dev-master@dev
     - env: SONATA_CORE=dev-master@dev
     - env: SONATA_ADMIN=dev-master@dev
+    - php: 7.1
+      env: SYMFONY=3.4.x-dev
```
greg0ire: We manage this file with https://github.com/sonata-project/dev-kit, you should not edit it.

Jan0707: Is there documentation ?

greg0ire: Regarding how dev-kit works, I'm afraid not, but you don't need to touch that file to get tests for sf4. Only for sf 3.4. Here is the file we will change when 3.4 is out: https://github.com/sonata-project/dev-kit/blob/master/config/projects.yml

core23:
Fixed with https://github.com/sonata-project/dev-kit/pull/351

PR #955, opened 2017-11-27
Ensure subject is an object in UserAclVoter

pierredup:
I am targeting this branch, because it is the lowest branch where the issue exists.

## Changelog

```markdown
### Fixed
- Added a check to the UserAclVoter class to ensure the subject is an object
```
## Subject

The `UserAclVoter` currently expects the subject to always be an object, but that isn't always the case. In the Symfony CMF ResourceRestBundle, subject is an array, which causes an error with the `get_class` method. So this adds a check to ensure the subject on the voter is an object before continuing with the rest of the checks

### DIFF
--- a/Security/Authorization/Voter/UserAclVoter.php
+++ b/Security/Authorization/Voter/UserAclVoter.php
@@ -41,7 +41,7 @@ class UserAclVoter extends AclVoter
      */
     public function vote(TokenInterface $token, $object, array $attributes)
     {
-        if (!$this->supportsClass(get_class($object))) {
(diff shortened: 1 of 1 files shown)


OskarStark:
Sorry, but `2.x` is not supported anymore and will not get a fresh release.
Could you please apply your fix against `3.x` if necessary?

Thank you

pierredup:
I originally did the fix against 3.x, but then saw the 2.x branch and wasn't sure which one to target. New PR at #956

greg0ire:
Great, but please be aware that you can rebase any PR with the "Edit" button, `git rebase` and `git push --force`

PR #978, opened 2017-12-08
Set choices_as_values to true in SecurityRolesType

EmmanuelVella:
## Changelog

```markdown
### Fixed
- Fixed flipped choices values/labels in SecurityRolesType when using symfony 2.8
```

## To do
- [x] Update the tests

## Subject

On the 4.x branch, `SecurityRolesType` choices are now flipped, but as this branch is compatible with symfony 2.8, we must set the `choices_as_values` option to `true` (if defined).

### DIFF
--- a/src/Form/Type/SecurityRolesType.php
+++ b/src/Form/Type/SecurityRolesType.php
@@ -133,6 +133,11 @@ class SecurityRolesType extends AbstractType
 
             'data_class' => null,
         ]);
+
+        // Symfony 2.8 BC
+        if ($resolver->isDefined('choices_as_values')) {
+            $resolver->setDefault('choices_as_values', true);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Form/Type/SecurityRolesType.php
```diff
@@ -133,6 +133,11 @@ public function configureOptions(OptionsResolver $resolver): void
 
             'data_class' => null,
         ]);
+
+        // Symfony 2.8 BC
+        if ($resolver->isDefined('choices_as_values')) {
```
greg0ire: OMG this is so much cleaner that all the other hacks!

EmmanuelVella:
Tests are now ok !

src/Form/Type/SecurityRolesType.php
OskarStark: love it 😻

PR #980, opened 2017-12-12
Fixed a bug with inability to use external model classes

### LINKED ISSUE #967: [4.0] way to disable checkManagerType or allow FOSUser as parent?
For one of my projects i used FOSUser as base class for my admin users because the Sonata users had too much Social Media stuff i didn't wanted as entries in the backend or the database.

I did wrote my own admin classes, extending the ones from Sonata, removed the Social Media stuff and it worked.

but now with checkManagerTypeToModelTypeMapping it doesn't work anymore and i need to extend from SonataUsers and do get the the facebook fields in the  database even if i don't want to.

could there be a way to disable the checkMangerType via a config flag, or allow FOSUser(Interface) / FOSGroup(Interface) to be a parent of that stuff?

vyshkant:

Closes #967

## Changelog

```markdown
### Fixed
- Fixed a bug with inability to use external model classes
```

## Subject

Existing consistency checks (SonataUserExtension::checkManagerTypeToModelTypeMapping) led to the inability to use external model class.

The changes consist in the fact that now only the classes inherited from `SonataUserBundle` base models are checked.

### DIFF
--- a/src/DependencyInjection/SonataUserExtension.php
+++ b/src/DependencyInjection/SonataUserExtension.php
@@ -87,7 +87,7 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
             $loader->load('security_acl.xml');
         }
 
-        $this->checkManagerTypeToModelTypeMapping($config);
+        $this->checkManagerTypeToModelTypesMapping($config);
 
         $this->registerDoctrineMapping($config);
         $this->configureAdminClass($config, $container);
@@ -261,45 +261,42 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
     /**
      * @param array $config
      */
-    private function checkManagerTypeToModelTypeMapping(array $config): void
+    private function checkManagerTypeToModelTypesMapping(array $config): void
     {
         $managerType = $config['manager_type'];
 
-        $actualModelClasses = [
-            $config['class']['user'],
-            $config['class']['group'],
-        ];
-
-        if ('orm' === $managerType) {
-            $expectedModelClasses = [
-                EntityUser::class,
-                EntityGroup::class,
-            ];
-        } elseif ('mongodb' === $managerType) {
-            $expectedModelClasses = [
-                DocumentUser::class,
-                DocumentGroup::class,
-            ];
-        } else {
+        if (!in_array($managerType, ['orm', 'mongodb'])) {
             throw new \InvalidArgumentException(sprintf('Invalid manager type "%s".', $managerType));
         }
 
-        foreach ($actualModelClasses as $index => $actualModelClass) {
-            if ('\\' === substr($actualModelClass, 0, 1)) {
-                $actualModelClass = substr($actualModelClass, 1);
-            }
+        $this->checkModelTypeMapping(
+            $config['class']['user'],
+            'orm' === $managerType ? DocumentUser::class : EntityUser::class,
+            $managerType
+        );
 
-            $expectedModelClass = $expectedModelClasses[$index];
+        $this->checkModelTypeMapping(
+            $config['class']['group'],
+            'orm' === $managerType ? DocumentGroup::class : EntityGroup::class,
+            $managerType
+        );
+    }
 
-            if ($actualModelClass !== $expectedModelClass && !is_subclass_of($actualModelClass, $expectedModelClass)) {
-                throw new \InvalidArgumentException(
-                    sprintf(
-                        'Model class "%s" does not correspond to manager type "%s".',
-                        $actualModelClass,
-                        $managerType
-                    )
-                );
-            }
+    /**
+     * @param string $actualModelClass
+     * @param string $prohibitedModelClass
+     * @param string $managerType
+     */
+    private function checkModelTypeMapping($actualModelClass, $prohibitedModelClass, $managerType): void
+    {
+        if ($actualModelClass === $prohibitedModelClass || is_subclass_of($actualModelClass, $prohibitedModelClass)) {
+            throw new \InvalidArgumentException(
+                sprintf(
+                    'Model class "%s" does not correspond to manager type "%s".',
+                    $actualModelClass,
+                    $managerType
+                )
+            );
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/DependencyInjection/SonataUserExtension.php
```diff
@@ -261,45 +261,47 @@ protected function aliasManagers(ContainerBuilder $container, $managerType): voi
...
-            if ($actualModelClass !== $expectedModelClass && !is_subclass_of($actualModelClass, $expectedModelClass)) {
-                throw new \InvalidArgumentException(
-                    sprintf(
-                        'Model class "%s" does not correspond to manager type "%s".',
-                        $actualModelClass,
-                        $managerType
-                    )
-                );
-            }
+    private function checkModelTypeMapping($managerType, $actualModelClass, $baseModelClass, $expectedModelClass = null)
```
core23: Please add a phpDoc

Hanmac:
@vyshkant it still isn't enough for me, i use FOS\UserBundle\Model\User as base for my user model class
so it should probably use that in the check too

vyshkant:
> @vyshkant it still isn't enough for me, i use FOS\UserBundle\Model\User as base for my user model class so it should probably use that in the check too

@Hanmac I excluded external classes from the check, so it should work when I fix it.

Hanmac:
@vyshkant i need to test your PR but it seems travis is still failing

for now, your checkModelTypeMapping would fail for me, because my Application\Sonata\UserBundle\Entity\User does extend from FOS\UserBundle\Model\User, not from Sonata\UserBundle\Model\User

currently, $expectedModelClass  is still checked in your code

vyshkant:
@Hanmac you can try now.

cocolabssas:
Some news?

jordisala1991:
@cocolabssas Have you tried the bundle with this PR?

cocolabssas:
@jordisala1991 Yes i replaced 4.0.0 version by this PR and it works. No need to have classes inherited from SonataUserBundle.

cocolabssas:
What are we waiting for to merge this PR?

vyshkant:
> Please, use the interfaces here to check against the configuration https://github.com/FriendsOfSymfony/FOSUserBundle/tree/master/Model

I'm not sure this is possible: now I can only check whether an `EntityModel` does (not) correspond to `MongodbOrmManager` and the like.

And there is no `EntityUserInterface` or `DocumentGroupInterface`, there are only `BaseUser` and `BaseGroup` classes in each section (orm and mongodb).

jordisala1991:
Why it is not possible to check against this interfaces?

https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Model/GroupInterface.php
https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Model/UserInterface.php

vyshkant:
> Why it is not possible to check against this interfaces?

Because these interfaces does not tell us anything about persistent level, which integrity I'm trying to check.

> All users and/or groups should extend this one

Exactly. And this is a problem, because in my check I'm trying to detect "model" and "orm manager" discrepancy, which cannot be done with `FOSUserBundle/Model/UserInterface`, because, as you said, "all users should extend this one".

jordisala1991:
Then I don't get what are you trying to check here.

IMHO, we should check if someone is trying to set a class for user that it cannot be a user, the same for the group.

vyshkant:
> Then I don't get what are you trying to check here.

I'm trying to check the correspondence between model and entity manager. To avoid the situation, when `Entity\BaseUser` is set together with `mongodb` manager type. Or vice versa: `Document\BaseUser` with `orm`.

Generally speaking, this check is already in the code, I'm just a little weakened its condition (it now applies only to the `SonataUserBundle\Model\User` children).

> IMHO, we should check if someone is trying to set a class for user that it cannot be a user, the same for the group.
:+1: but that's not what I did.

Hanmac:
I don't know if you can access the services yet at this place, but using the main 'doctrine' Service and then use 'getManagerForClass' and then check if that Manager would have the wanted type or something

Hanmac:
hm i think for a later change we might rethink the use of the user manager & group manager and if we really need different versions for them in entity & document or if we could make a generic one that does that somewhere below in the code?

jeroendk:
Any updates on this ?

vyshkant:
> Any updates on this ?

@core23 @jordisala1991

peter-gribanov:
Maybe it's worth inverting the nonconformity check?
If manager type is `orm` then User should not extends `Document\BaseUser`
If manager type is `mongodb` then User should not extends `Entity\BaseUser`

There should not be strict restrictions.
This is a violation of ISP

triion:
**Any updates here?** This strict restriction is blocking a clean User-Entity. 

Would it be an alternative to add a BareUser-Model (+BareGroup) to Sonata UserBundle (empty copy of FOS/User?) that can always be used as Parent to extend the User and Group Entities as seen fit by the specific projects... BaseUser can extend BareUser and all Entities can be used for the nonconformity check...

Now we are "stuck" with many columns in User that we don't really need, pulluting our DB...

jordisala1991:
IMO, as I said before:

I would prefer to get the restriction on the lowest possible level:

For user: https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Model/UserInterface.php
For group: https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Model/GroupInterface.php

No need to add restrictions for ORM, ODM, etc..

klipach:
Are there any updates for this topic? It is still not possible to use custom User/Group classes that are implementing GroupInterface/UserInterface.

core23:
Please rebase and apply these changes: https://github.com/sonata-project/SonataUserBundle/pull/980#issuecomment-381893326

jeroendk:
@vyshkant Any plans to finish this one? Could I be of any assistance?

javer:
@vyshkant Could you please apply required changes to make this PR mergeable? Because it blocks as to use any newer version than 505de3e9fdaf096a9d4b3512bba2d43364992b42, which is very old at the moment. Maybe I or someone else could help you?

OskarStark:
Please provide this as a new PullRequest, thank you

jordisala1991:
So this one should be closed right? via #1095

PR #996, opened 2018-01-13
[Bug] Gender list - flip values

mimol91:
I am targeting this branch, because o there **is BC-break**

## Subject

There was an change in BaseUser.php https://github.com/sonata-project/SonataUserBundle/commit/c69d51bdd72c50340aa04c9b14a95b18e8600336#diff-0b70ecd19ebc26172322781dd1061c2fL47 which caused flip optionValues with optionLabel. - It generate error because of extending `https://github.com/sonata-project/SonataCoreBundle/blob/master/src/Form/Type/BaseStatusType.php` which set `flip` to true.

It could be updated aswell by adding fourth argument to `https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/Resources/config/form.xml#L7` set to `false`, however as mentioned in comment it will be removed in `v4`

## Changelog

```markdown
### Fixed
Fix problem with flipped values in `getGenderList` method to match expected arguments in `BaseStatusType`
```

### DIFF
--- a/src/Entity/BaseUser.php
+++ b/src/Entity/BaseUser.php
@@ -46,9 +46,9 @@ class BaseUser extends AbstractedUser
     public static function getGenderList()
     {
         return [
-            'gender_unknown' => UserInterface::GENDER_UNKNOWN,
-            'gender_female' => UserInterface::GENDER_FEMALE,
-            'gender_male' => UserInterface::GENDER_MALE,
+            UserInterface::GENDER_UNKNOWN => 'gender_unknown',
+            UserInterface::GENDER_FEMALE => 'gender_female',
+            UserInterface::GENDER_MALE => 'gender_male',
         ];
     }
 }


greg0ire:
Hey there!

It looks like you did not fill the pull request template properly. Maybe you
were confused, so let me walk you through this step by step.

# Explaining why you target this branch

> I am targeting this branch, because {reason}.

Here, you are supposed to replace `{reason}` with something like:

- "there is a BC-break" if you picked the master branch
- "there is no BC-break" if you picked the stable branch

# Filling the changelog

That one is really important to get right, because it will be used to craft a
release changelog such as [this
one](https://github.com/sonata-project/SonataAdminBundle/releases/tag/3.23.0).
The first thing you need to know about this is that if your pull request is
pedantic, meaning it does not fix a bug for the user, or adds a new feature, it
is not needed. This is typically the case for PRs that fix the build or add
some documentation.

3 important things:

1. Make sure to fill the changelog section with something directed at the end
user, that answers the question "What benefit do I get from upgrading?".
2. Make sure to remove empty sections.
3. Make sure to keep the markdown markup.

greg0ire:
Flipping those might indeed have been a bad idea (or at least, a BC-break). @jordisala1991 please advise

src/Entity/BaseUser.php
```diff
@@ -46,9 +46,9 @@ public function preUpdate(): void
     public static function getGenderList()
     {
         return [
-            'gender_unknown' => UserInterface::GENDER_UNKNOWN,
-            'gender_female' => UserInterface::GENDER_FEMALE,
-            'gender_male' => UserInterface::GENDER_MALE,
+            UserInterface::GENDER_UNKNOWN => 'gender_unknown',
```
core23: This is a BC break. You should flip this values in your form type

greg0ire: The implication is that this was a BC break too, right: https://github.com/sonata-project/SonataUserBundle/commit/c69d51bdd72c50340aa04c9b14a95b18e8600336#diff-0b70ecd19ebc26172322781dd1061c2fL47 ?

greg0ire: Oh but it was in a major PR, so, completely normal.

mimol91: I was thinking about BC,, however w/o this change, it seems that it does not work (at least on 4.0.2), so I was created this simple fix.
I've extended this form type, and flipped it myself, however I think it should be in upcoming version (otherwise you will get some ugly sql error  - gender type is varchar(1) -> and it try to save `gender_unknown`

core23: > it seems that it does not work (at least on 4.0.2), so I was created this simple fix.

The problem is not the method in the entity itself. Your problem is probably in your FormType

mimol91: @core23
It caused because of https://github.com/sonata-project/SonataCoreBundle/blob/master/src/Form/Type/BaseStatusType.php
with combination with https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/Resources/config/form.xml#L7
and 
https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/Form/Type/UserGenderListType.php#L18

`UserGenderListType` extends `BaseStatusType` which unfortunatelly has 4th argument set to `null` which cause that values returned from `getGenderList` method are flipped

jordisala1991: what about tell the BaseStatusType to not flip those choices?

mimol91: it was my 1st idea, however It says that 4th argument is depreciated, so it has to be changed again in v4

jordisala1991: @core23 what we should do here? If we cannot override the fourth parameter (deprecated) we need to flip the choices right?

core23: Create a new `AbstractStatusType`? https://github.com/sonata-project/SonataCoreBundle/pull/499#issuecomment-357534483

core23: For this issue, we must keep this mapping and can't flip this right now. Please change the branch for this PR to master an leave an upgrade note.

mimol91: Should I add note to `UPGRADE-4.x.md` or is PR description itself?

OskarStark:
@silasjoisten this could be interesting for you!

src/Entity/BaseUser.php
jordisala1991: IMO it is a better solution to create a new AbstactStatus time. And even better, why do we need a type that is only extending the choiceType. Can you try instead changing the admin to use ChoiceType directly? The choices are on  a static method, so you can access them.

laffet:
i was able to solve this problem by overiding the method getGenderList() in my user entity that extends BaseUser (Sonata\UserBundle\Entity\BaseUser)
capture d ecran de 2019-02-08 11-11-20

PR #1007, opened 2018-01-27
Fix gender choices flip in user sonata form

### LINKED ISSUE #992: Translation and registration database errors on Sonata\UserBundle\Entity\BaseUser
There is an error in the ````Sonata\UserBundle\Entity\BaseUser````. There is a reversal of key values ​​that causes translation problems and prevents the correct backup of the "gender" field

```php
    // v3.* working successfully 
    public static function getGenderList()
    {
        return [
            UserInterface::GENDER_UNKNOWN => 'gender_unknown',
            UserInterface::GENDER_FEMALE => 'gender_female',
            UserInterface::GENDER_MALE => 'gender_male',
        ];
    }

    // v4.* does not work
    public static function getGenderList()
    {
        return [
            'gender_unknown' => UserInterface::GENDER_UNKNOWN,
            'gender_female' => UserInterface::GENDER_FEMALE,
            'gender_male' => UserInterface::GENDER_MALE,
        ];
    }
```

jordisala1991:

Closes #992 
Closes #996 

## Changelog

```markdown
### Fixed
- choices for User gender now appears correctly flipped and translated
```

## Todo 

- [x] Add an upgrade note

## Subject

BaseStatusType is just a ChoiceType. IMO let's try to move everything to choice type and deprecate that BaseStatusType someday. Other bundles using this: ecommerce and comment.

Without this patch it is no possible to safely save an user to the database using the sonata admin, because it will try to save something like `gender_male` to a column that only accepts a single char string


src/Form/Type/UserGenderListType.php
```diff
@@ -15,6 +15,17 @@
...
+@trigger_error(
+    'The '.__NAMESPACE__.'\UserGenderListType class is deprecated since version 4.x and will be removed in 5.0.'
+    .' Use choice type instead.',
+    E_USER_DEPRECATED
+);
+
+/**
+ * NEXT_MAJOR: remove this class.
+ *
+ * @deprecated since version 4.x, to be removed in 5.0. Use choice type instead
```
core23: Why did you deprecate this?

Couldn't you move the stuff from the `UserAdmin` right here?

jordisala1991: why would I use a class to solve this simple choice thing?

jordisala1991:

src/Form/Type/UserGenderListType.php
core23: The problem is, that the `UserGenderListType` is broken, not the admin class (only)

jordisala1991: Is broken and cannot be fixed unless we allow deprecations (adding the fourth parameter to false) or more BC breaks. And it is because of this other PR: https://github.com/sonata-project/SonataCoreBundle/pull/499 which breaks BC.

My first thought was to implement a new AbstractStatusType but I there are 2 reasons why I didn't do it:

1. I would need to change which class is extending UserGenderListType (probably BC break too)
2. This can be solved in a much easier way, without using this class, and using a simple choiceType.

How to proceed? At this moment the user admin class is unusable

core23: Can't you just add this to the form type?

```
        // NEXT_MAJOR: Remove this when dropping support for SF 2.8
        if (method_exists(FormTypeInterface::class, 'setDefaultOptions')) {
            $genderOptions['choices_as_values'] = true;
        }
```

UPGRADE-4.x.md
```diff
@@ -1,2 +1,7 @@
 UPGRADE 4.x
 ===========
+
+### Deprecated UserGenderListType
+
+Relying on this class to provide user gender list on sonata forms is deprecated and
+will be removed in 5.0. You should use choice type instead.
```
greg0ire: Either "the choice type" or `ChoiceType` (or even `Symfony\Component\Form\Extension\Core\Type\ChoiceType`)

jordisala1991: Fixed

composer.json
```diff
@@ -29,9 +29,14 @@
...
         "sonata-project/easy-extends-bundle": "^2.2",
         "swiftmailer/swiftmailer": "^4.3 || ^5.0 || ^6.0",
+        "symfony/config": "^2.8 || ^3.2 || ^4.0",
         "symfony/console": "^2.8 || ^3.2 || ^4.0",
+        "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
         "symfony/form": "^2.8 || ^3.2 || ^4.0",
+        "symfony/framework-bundle": "^2.8 || ^3.2 || ^4.0",
         "symfony/http-foundation": "^2.8 || ^3.2 || ^4.0",
+        "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
+        "symfony/options-resolver": "^2.8 || ^3.2 || ^4.0",
```
kunicmarko20: Shouldn't this be in a different PR as it is not connected to the subject?

jordisala1991: It is somewhat related because the build was failing on prefer-lowest because we were using older versions of this deps. Once I explicitly declare this dependencies on composer.json it works again.

src/Form/Type/UserGenderListType.php
```diff
@@ -15,6 +15,17 @@
 
 use Sonata\CoreBundle\Form\Type\BaseStatusType;
 
+@trigger_error(
+    'The '.__NAMESPACE__.'\UserGenderListType class is deprecated since version 4.x and will be removed in 5.0.'
+    .' Use ChoiceType instead.',
```
OskarStark: Please use FQCN

src/Form/Type/UserGenderListType.php
```diff
@@ -15,6 +15,17 @@
...
+    E_USER_DEPRECATED
+);
+
+/**
+ * NEXT_MAJOR: remove this class.
+ *
+ * @deprecated since version 4.x, to be removed in 5.0. Use ChoiceType instead
```
OskarStark: Please use FQCN

src/Resources/config/form.xml
```diff
@@ -2,6 +2,7 @@
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="sonata.user.form.gender_list" class="Sonata\UserBundle\Form\Type\UserGenderListType">
+            <deprecated>The "%service_id%" service is deprecated since 4.x and will be removed in 5.0. Use choice type instead</deprecated>
```
OskarStark: Please use FQCN for choice type

src/Form/Type/UserGenderListType.php
jordisala1991: Fixed

src/Form/Type/UserGenderListType.php
jordisala1991: Fixed

src/Resources/config/form.xml
jordisala1991: Fixed, not using the full message since symfony does the same for its classes: 

https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Security/Core/SecurityContext.php#L14

vs

https://github.com/symfony/symfony/blob/2.8/src/Symfony/Bundle/SecurityBundle/Resources/config/security.xml#L57

Hope you agree

jordisala1991:
Added test too, for the deprecation. Please review again @OskarStark @core23 and @greg0ire

jordisala1991:
Still waiting for your approval @core23

core23:
For me, this is just a hotfix for the local issue (inside the admin class). This will not fix any problem with `UserGenderListType` itself. If you use this class inside a custom user form, you got still got the bug. 

I'm okay if we deprecate `UserGenderListType`, but the bug inside this class should be fixed too if you reply on it (outside the admin class).

jordisala1991:
Fix the “issue” if you want in another PR but do not block this one. IMO the issue is fixed the moment we say this class is deprecated and should not be used

If you are okay with this one, please approve

PR #1008, opened 2018-01-29
Fix choices_as_values deprecation message

jordisala1991:

## Changelog

```markdown
### Fixed
- Deprecation message on SecurityRolesType about `choices_as_values`
```

## Subject

Seems like the deprecations build is failing because of this.
Link: https://travis-ci.org/sonata-project/SonataUserBundle/jobs/334884037#L798

Probably this makes this bundle unusable with SF 4, that's why I am tagging it with patch

### DIFF
--- a/src/Form/Type/SecurityRolesType.php
+++ b/src/Form/Type/SecurityRolesType.php
@@ -16,10 +16,12 @@ namespace Sonata\UserBundle\Form\Type;
 use Sonata\UserBundle\Form\Transformer\RestoreRolesTransformer;
 use Sonata\UserBundle\Security\EditableRolesBuilder;
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormEvent;
 use Symfony\Component\Form\FormEvents;
 use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormTypeInterface;
 use Symfony\Component\Form\FormView;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
@@ -135,7 +137,7 @@ class SecurityRolesType extends AbstractType
         ]);
 
         // Symfony 2.8 BC
-        if ($resolver->isDefined('choices_as_values')) {
+        if (method_exists(FormTypeInterface::class, 'setDefaultOptions')) {
             $resolver->setDefault('choices_as_values', true);
         }
     }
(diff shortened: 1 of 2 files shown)


src/Form/Type/SecurityRolesType.php
```diff
@@ -135,7 +137,7 @@ public function configureOptions(OptionsResolver $resolver): void
         ]);
 
         // Symfony 2.8 BC
-        if ($resolver->isDefined('choices_as_values')) {
+        if (method_exists(FormTypeInterface::class, 'setDefaultOptions')) {
```
greg0ire: Can you explain the fix?

tests/Form/Type/SecurityRolesTypeTest.php
```diff
@@ -94,25 +98,6 @@ public function testSubmitWithHiddenRoleData(): void
         $this->assertContains('ROLE_SUPER_ADMIN', $form->getData());
     }
 
-    public function testChoicesAsValues(): void
```
greg0ire: Why is this removed?

src/Form/Type/SecurityRolesType.php
kunicmarko20: choices_as_values is removed in Symfony 4

greg0ire: Which means `$resolver->isDefined('choices_as_values')` will return `false`, which means `$resolver->setDefault('choices_as_values', true);` will not be set to `true`, which… makes sense if that option is removed?

kunicmarko20: The default value of `choices_as_values` is false and that shows deprecation notices, but we do not have that option in SF4, so he checks if it is < SF3 and sets it to true to avoid deprecation notices. Am I correct @jordisala1991 ?

jordisala1991: Yes, that's also probably. I just wanted to remove the deprecations on Sf3, but wasn't sure if this would work on Sf4

tests/Form/Type/SecurityRolesTypeTest.php
jordisala1991: Because this is tested already on the test above, but conditionally only if there are choices_as_values option

src/Form/Type/SecurityRolesType.php
```diff
@@ -135,7 +137,7 @@ public function configureOptions(OptionsResolver $resolver): void
...
             $resolver->setDefault('choices_as_values', true);
```
greg0ire: The coverage decreased and coveralls says this is no longer covered by tests… is that true and normal?

kunicmarko20: It is covered on 2.8 https://coveralls.io/jobs/33261397/source_files/324132516#L140, because `setDefaultOptions` exist only there.

greg0ire:
Thanks @jordisala1991 !

PR #1009, opened 2018-01-30
Fix gender choices type

core23:
I am targeting this branch, because this is a bugfix.

## Changelog

```markdown
### Fixed
- choices for User gender now appears correctly flipped and translated
```

## Subject

This is a bugfix for #1007, so the `UserGenderListType` will work again in symfony 3.x.

### DIFF
--- a/src/Form/Type/UserGenderListType.php
+++ b/src/Form/Type/UserGenderListType.php
@@ -14,19 +14,32 @@ declare(strict_types=1);
 namespace Sonata\UserBundle\Form\Type;
 
 use Sonata\CoreBundle\Form\Type\BaseStatusType;
+use Symfony\Component\Form\FormTypeInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
-@trigger_error(
-    'The '.__NAMESPACE__.'\UserGenderListType class is deprecated since version 4.1 and will be removed in 5.0.'
-    .' Use Symfony\Component\Form\Extension\Core\Type\ChoiceType instead.',
-    E_USER_DEPRECATED
-);
-
-/**
- * NEXT_MAJOR: remove this class.
- *
- * @deprecated since version 4.1, to be removed in 5.0.
- * Use Symfony\Component\Form\Extension\Core\Type\ChoiceType instead
- */
 class UserGenderListType extends BaseStatusType
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        // NEXT_MAJOR: Call the parent to set the choices
+        $choices = \call_user_func([$this->class, $this->getter]);
+
+        // Only flip choice list, if the keys are m/f/u and not labels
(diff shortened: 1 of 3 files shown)


jordisala1991:
IMHO having a class extending another that will be deprecated soon ( because of the wrong naming and incorrect flip ) and having to duplicate code just because you dont want to deprecate it, is a waste

core23:
If you don't want to use this class anymore, we should deprecate `BaseStatusType`  too

jordisala1991:
And I will , after I replace all uses and deprecate all clases extending it

jordisala1991:
IMO if having an extra choice type has any benefits lets go for it. But it does not, because it adds nothing but problems with version compatibility

src/Form/Type/UserGenderListType.php
```diff
@@ -29,4 +31,22 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        // NEXT_MAJOR: Call the parent to set the choices
+        $choices = call_user_func([$this->class, $this->getter]);
+
+        $resolver->setDefaults([
+            'choices' => array_flip($choices),
```
jordisala1991: In theory you dont need to flip them again

core23: Made some adjustments to detect if we need a flip or not.

src/Form/Type/UserGenderListType.php
```diff
@@ -29,4 +31,27 @@
  */
 class UserGenderListType extends BaseStatusType
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        // NEXT_MAJOR: Call the parent to set the choices
```
jordisala1991: You can safely remove this NEXT_MAJOR, cause this class is gonna be removed on 5.0

src/Form/Type/UserGenderListType.php
```diff
@@ -29,4 +31,27 @@
...
+        $choices = call_user_func([$this->class, $this->getter]);
+
+        // Only flip choice list, if the keys are labels and not actual keys
+        if (1 !== strlen(key($choices))) {
+            $choices = array_flip($choices);
+        }
+
+        $resolver->setDefaults([
+            'choices' => $choices,
+            'choice_translation_domain' => 'SoantaUserBundle',
```
jordisala1991: Shouldn't this come from the admin class?

greg0ire: Should be `Sonata`, not `Soanta`

OskarStark:
could you finalize this PR @core23 ?

OskarStark:
@core23 any interests on this PR? If not please close, thanks

core23:
Done @OskarStark

greg0ire:
The build failed

OskarStark:
Can you please rebase and fix Travis? This should be merged

PR #1013, opened 2018-02-14
Remove support for FOSRest <2.1

jordisala1991:

## Changelog

```markdown
### Removed
- Removed compatibility with older versions of FOSRestBundle (<2.1)
```

## Subject

Since dropping a dependency is not a BC break, this is a good start, because we are doing a lot of code just to keep compat with FOSRest 1.

FOSUserBundle is the only one blocking SF 4 here now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "symfony/translation": "^2.8 || ^3.2 || ^4.0"
     },
     "conflict": {
-        "friendsofsymfony/rest-bundle": "<1.1",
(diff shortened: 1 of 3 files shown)


jordisala1991:
I can create another PR for all bundles if we need to change that.

I found that there is some code that could be improved on the apis, so I am probably going for a second round of PR for all of them anyway

greg0ire:
Thanks @jordisala1991 !

PR #1019, opened 2018-03-30
Fix locale orm db field length

tifabien:
I am targeting this branch, because it fixes the locale db field length and it's considered as a BC break.

## Changelog
```markdown
### Fixed
 - Fixed locale db field length
```
## Subject
When we try to create a user with the 'sr_Latn_RS' it fails due to the locale db field length.

### DIFF
--- a/src/Resources/config/doctrine/BaseUser.orm.xml
+++ b/src/Resources/config/doctrine/BaseUser.orm.xml
@@ -10,7 +10,7 @@
         <field name="website" type="string" column="website" length="64" nullable="true"/>
         <field name="biography" type="string" column="biography" length="1000" nullable="true"/>
         <field name="gender" type="string" column="gender" length="1" nullable="true"/>
-        <field name="locale" type="string" column="locale" length="8" nullable="true"/>
+        <field name="locale" type="string" column="locale" length="15" nullable="true"/>
         <field name="timezone" type="string" column="timezone" length="64" nullable="true"/>
         <field name="phone" type="string" column="phone" length="64" nullable="true"/>
         <!-- social fields -->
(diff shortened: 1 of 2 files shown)


src/Resources/config/doctrine/BaseUser.orm.xml
```diff
@@ -10,7 +10,7 @@
         <field name="website" type="string" column="website" length="64" nullable="true"/>
         <field name="biography" type="string" column="biography" length="1000" nullable="true"/>
         <field name="gender" type="string" column="gender" length="1" nullable="true"/>
-        <field name="locale" type="string" column="locale" length="8" nullable="true"/>
+        <field name="locale" type="string" column="locale" length="10" nullable="true"/>
```
kunicmarko20: Isn't this a BC break? I think it should be targeted on master

kunicmarko20: Well, if we just change it here everyone using sonata user would get a database error. You can still do http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/override-field-association-mappings-in-subclasses.html. Let's see and wait for others. ping @sonata-project/contributors

jordisala1991: If we are changing this it would be nice to give it a few extra chars, to not having this problem again. And probably change it on master to avoid bc breaks

tifabien: OK for me. Updated the base branch and increased to 15 the length

kunicmarko20:
I think we don't need upgrade note here, so I guess nothing. cc @sonata-project/contributors

greg0ire:
> I think we don't need upgrade note here, so I guess nothing. cc @sonata-project/contributors

If there's a BC-break, then you need an upgrade note. In this case, it should say "make sure you run your migrations" I guess.

kunicmarko20:
So I am wrong, do what @greg0ire said 😄

greg0ire:
Thanks @tifabien !

PR #1020, opened 2018-03-30
Fix jms config in docs

kunicmarko20:
I am targeting this branch, because docs.

## Subject

A name is required when defining paths.


greg0ire:
Thanks @kunicmarko20!

PR #1022, opened 2018-04-16
Removed redundant "required" asterisk

BitScout:
## Changelog

```markdown

### Changed
- Removed redundant "required" asterisk
```

## Subject

The label contains two asterisks: One hard-coded and one from the class "required", which includes an asterisk in the :after.

### DIFF
--- a/src/Resources/views/Resetting/request_content.html.twig
+++ b/src/Resources/views/Resetting/request_content.html.twig
@@ -15,7 +15,7 @@ file that was distributed with this source code.
     {% endif %}
 
     <div class="form-group">
-        <label for="username" class="control-label required">{{ 'resetting.request.username'|trans({}, 'FOSUserBundle') }} * </label>
+        <label for="username" class="control-label required">{{ 'resetting.request.username'|trans({}, 'FOSUserBundle') }}</label>
         <input type="text" id="username" name="username" required="required" />
     </div>
 


qsomazzi:
Hello @sonata-project/contributors .
This is something we've discover on our project, here's a screen to provide more visibility.

capture d ecran 2018-04-16 a 15 57 41

kunicmarko20:
Branch 3.x is not maintained anymore, please use 4.x. If this is still a problem on 4.x change branch of this PR.

qsomazzi:
we use SonataUser in 3.x, so we need it to be fixed on this branch, not 4.x.

Soullivaneuh:
Sorry @qsomazzi, we won't merge this PR for these reasons:

- Because the 4.0 release is here from December 2017 and we are supposed to maintain only one stable branch.
- Because for this case, the hard-coded asterisk is maybe here for a reason and it would be great to not have "fixes of fixes" on a legacy branch.
- AFAIK, you can override the login template with ease waiting the 4.0 upgrade.

greg0ire:
>  i don't get the issue to fix a not maintained version

"Not maintained" literally means we don't even try to fix it, or to merge PRs that do, or make releases that include said PRs

kunicmarko20:
As I don't see this in 4.x:

https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/Resources/views/Admin/Security/Resetting/request.html.twig

can we close this?

BitScout:
Well, that's going far faster than expected. Bummer that there's a fixed date where support was just switched from 3.x to 4.x. Hard to work with that.

kunicmarko20:
We are sorry about that, but that is why you should upgrade. We are a team of few people and we can't maintain everything. You have to understand us too. 🙂

greg0ire:
> Bummer that there's a fixed date where support was just switched from 3.x to 4.x.

Sorry, that's how we decided to use our resources, which are scarce.

PR #1025, opened 2018-05-08
4.2.0

greg0ire:
Please just approve, don't merge

### DIFF
--- a/src/Form/Type/UserGenderListType.php
+++ b/src/Form/Type/UserGenderListType.php
@@ -16,7 +16,7 @@ namespace Sonata\UserBundle\Form\Type;
 use Sonata\CoreBundle\Form\Type\BaseStatusType;
 
 @trigger_error(
-    'The '.__NAMESPACE__.'\UserGenderListType class is deprecated since version 4.x and will be removed in 5.0.'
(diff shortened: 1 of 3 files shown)


src/Form/Type/UserGenderListType.php
```diff
@@ -16,15 +16,15 @@
 use Sonata\CoreBundle\Form\Type\BaseStatusType;
 
 @trigger_error(
-    'The '.__NAMESPACE__.'\UserGenderListType class is deprecated since version 4.x and will be removed in 5.0.'
+    'The '.__NAMESPACE__.'\UserGenderListType class is deprecated since version 4.1 and will be removed in 5.0.'
```
kunicmarko20: shouldn't it be 4.2?

kunicmarko20: same for rest

OskarStark: yes thought the same after approval

greg0ire: Apparently, this shipped with 4.1.1, hence the number.

greg0ire: See ad328649992b747e2a59bdb52bafd8c6bb90befd

PR #1038, opened 2018-07-03
fix(authenticator): fix namespace from sonata-project/GoogleAuthenticator

bmaziere:
I am targeting this branch, because i don't think it has BC break (i'm not sure because of https://github.com/sonata-project/SonataUserBundle/blob/e0eb6e0c7290f93b5e3ef850483233f7e957c454/composer.json#L61).

## Changelog
```markdown
### Fixed
Replace deprecated use of Google\Authenticator\GoogleAuthenticator by Sonata's namespace
```

## Subject
Calls to Google\Authenticator\GoogleAuthenticator raised errors when using composer with --classmap-authoritative option. 
Not completely sure but it might conflict with version 1 of sonata-project/google-authenticator as this namespace was only introduced in v2 (https://github.com/sonata-project/GoogleAuthenticator/commit/a339606c1ca9e8700ad74e91a8f9ff09bfa526db)

### DIFF
--- a/src/DependencyInjection/SonataUserExtension.php
+++ b/src/DependencyInjection/SonataUserExtension.php
@@ -67,6 +67,14 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
         $loader->load('form.xml');
 
         if (class_exists('Google\Authenticator\GoogleAuthenticator')) {
+            @trigger_error(
+                'The \'Google\Authenticator\' namespace is deprecated in sonata-project/GoogleAuthenticator since version 2.1 and will be removed in 3.0.',
+                E_USER_DEPRECATED
+            );
+        }
+
+        if (class_exists('Google\Authenticator\GoogleAuthenticator') ||
+            class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator')) {
             $loader->load('google_authenticator.xml');
         }
 
@@ -156,7 +164,8 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
             return;
         }
 
-        if (!class_exists('Google\Authenticator\GoogleAuthenticator')) {
+        if (!class_exists('Google\Authenticator\GoogleAuthenticator')
+            && !class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator')) {
             throw new \RuntimeException('Please add ``sonata-project/google-authenticator`` package');
         }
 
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/SonataUserExtension.php
```diff
@@ -66,7 +66,7 @@ public function load(array $configs, ContainerBuilder $container): void
 
         $loader->load('form.xml');
 
-        if (class_exists('Google\Authenticator\GoogleAuthenticator')) {
+        if (class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator')) {
```
kunicmarko20: I think this is a BC break, maybe if you do, `if (class_exists('Google\Authenticator\GoogleAuthenticator') || class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator'))`

kunicmarko20: or add a conflict to composer.json about the older version of library.

bmaziere: not sure that adding conflict on v1 will be a good solution as it may block installation & require php7 no? think `if (class_exists('Google\Authenticator\GoogleAuthenticator') || class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator'))` is better with a message to remove the `Google` root one on next major, WDYT?

kunicmarko20: the best way would be to add deprecations on stable and fix on master.

bmaziere: done for this one @kunicmarko20 . will give a try to fix it on master

kunicmarko20:
Could you please add deprecation message and connect commit to your github account/email

bmaziere:
done @kunicmarko20 ;)

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -66,7 +66,10 @@ public function load(array $configs, ContainerBuilder $container): void
 
         $loader->load('form.xml');
 
-        if (class_exists('Google\Authenticator\GoogleAuthenticator')) {
+        // The 'Google\Authenticator' namespace is deprecated in sonata-project/GoogleAuthenticator
+        // since version 2.1 and will be removed in 3.0.'
```
kunicmarko20: I meant throw deprecation from php
```
@trigger_error(
    'The 'Google\Authenticator' namespace is deprecated in sonata-project/GoogleAuthenticator since version 2.1 and will be removed in 3.0.',
    E_USER_DEPRECATED
);

bmaziere: if i do that the build can't pass and it's normal IMO. I already did this. I did it again so you can see

bmaziere: thx for the link ;) my bad i'm pretty new to deprecate things. i tried so tell me if it's ok

OskarStark:
Than you @bmaziere 🤞🏻

OskarStark:
4.2.3 is released

PR #1041, opened 2018-07-19
Fix google authenticator uses

bmaziere:
I am targeting this branch, because it is BC break. It is the following of https://github.com/sonata-project/SonataUserBundle/pull/1038

## Changelog

```markdown
### Changed
- Upgrade requirement of sonata-project/google-authenticator to 2.1
```

## Subject

Upgrading sonata-project/google-authenticator dependency to 2.1 and change namespace uses of GoogleAuthenticator

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,6 @@
         "nelmio/api-doc-bundle": "<2.4",
         "sonata-project/block-bundle": "<3.11",
         "sonata-project/doctrine-orm-admin-bundle": "<3.0",
-        "sonata-project/google-authenticator": "<1.0",
         "sonata-project/seo-bundle": "<2.0"
     },
     "require-dev": {
@@ -58,7 +57,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^1.1",
         "nelmio/api-doc-bundle": "^2.4",
         "sonata-project/block-bundle": "^3.11",
-        "sonata-project/google-authenticator": "^1.0 || ^2.0",
+        "sonata-project/google-authenticator": "^2.1",
         "sonata-project/seo-bundle": "^2.0",
         "symfony/phpunit-bridge": "^4.0"
     },
--- a/src/DependencyInjection/SonataUserExtension.php
+++ b/src/DependencyInjection/SonataUserExtension.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sonata\UserBundle\DependencyInjection;
 
 use Sonata\EasyExtendsBundle\Mapper\DoctrineCollector;
+use Sonata\GoogleAuthenticator\GoogleAuthenticator;
 use Sonata\UserBundle\Document\BaseGroup as DocumentGroup;
 use Sonata\UserBundle\Document\BaseUser as DocumentUser;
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -47,7 +47,7 @@
         "nelmio/api-doc-bundle": "<2.4",
         "sonata-project/block-bundle": "<3.11",
         "sonata-project/doctrine-orm-admin-bundle": "<3.0",
-        "sonata-project/google-authenticator": "<1.0",
+        "sonata-project/google-authenticator": "<2.1",
```
OskarStark: do we still need this conflict section entry?

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -66,15 +67,7 @@ public function load(array $configs, ContainerBuilder $container): void
...
-            @trigger_error(
-                'The \'Google\Authenticator\' namespace is deprecated in sonata-project/GoogleAuthenticator since version 2.1 and will be removed in 3.0.',
-                E_USER_DEPRECATED
-            );
-        }
-
-        if (class_exists('Google\Authenticator\GoogleAuthenticator') ||
-            class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator')) {
+        if (class_exists(GoogleAuthenticator::class)) {
             $loader->load('google_authenticator.xml');
```
OskarStark: couldn't we do the method call `$this->configureGoogleAuthenticator($config, $container);` here instead of line 105 and remove the second check for the class inside the method itself?

composer.json
bmaziere: true ;) removed

src/DependencyInjection/SonataUserExtension.php
bmaziere: not sure it is the good way as it is a require-dev dependency then i think first part of `configureGoogleAuthenticator` method has to be ran even if dependency is not installed on no dev context. WDYT?

bmaziere:
build fixed @OskarStark ;)

bmaziere:
should i try to fix the coveralls or is it enough to be merged?

OskarStark:
> should i try to fix the coveralls or is it enough to be merged?

It is fine, tank you!

tests/DependencyInjection/SonataUserExtensionTest.php
```diff
@@ -55,6 +51,31 @@ public function testLoadDefault(): void
...
+    public function testFixImpersonatingWithImpersonatingConfig(array $expectedConfig, array $providedConfig): void
+    {
+        $extension = new SonataUserExtension();
+
+        $this->assertSame($expectedConfig, $extension->fixImpersonating($providedConfig));
+    }
+
+    public function fixImpersonatingDataProvider(): \Generator
+    {
+        yield [['impersonating' => ['route' => 'foo', 'parameters' => []]], ['impersonating' => ['route' => 'foo', 'parameters' => []]]];
```
greg0ire: please add a key describing each case : `yield 'some description' =>`

bmaziere: done @greg0ire

PR #1044, opened 2018-08-29
Add invalid_username in response when username not found in sendEmailAction

morinceline:
## Changelog

```markdown
### Fixed
 - A case when user was not found in sendEmailAction
```
## Subject
On reset password, in case the user does not exist, then the controller will not work as expected. To fix it I add a test on `$user` variable and render the original template with `invalid_username` defined.

### DIFF
--- a/tests/Controller/AdminResettingControllerTest.php
+++ b/tests/Controller/AdminResettingControllerTest.php
@@ -21,6 +21,7 @@ use Symfony\Bundle\TwigBundle\TwigEngine;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Kernel;
 
 class AdminResettingControllerTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


kunicmarko20:
phpcsfixer style changes should be done in a different PR as it is not related to this PR.

kunicmarko20:
Blocked by #1045

morinceline:
@kunicmarko20 I have connect my email/github account and delete fix phpcsfixer ;)

kunicmarko20:
Thank you! Let's just wait for someone to merge the other PR and we can merge this too.

morinceline:
@kunicmarko20 4.x or master to target branch ?

kunicmarko20:
What is the current behavior if the user is `null`? Is the exception thrown? Also can you please add a test case

morinceline:
Currently if the user is null the behavior is the same as if the user exists. There is no exception, but in the twig there is a variable invalid_username to indicate that the user does not exist.

OK for tests.

kunicmarko20:
Then you can add it to `4.x` as it is a missing feature.

kunicmarko20:
Where did the rest of the file changes come in the pr?

morinceline:
it's an error to file changes ;)

morinceline:
@kunicmarko20 I added tests

kunicmarko20:
Travis is failing

morinceline:
I don't understand !! on my dev local the tests pass

tests/Controller/AdminResettingControllerTest.php
```diff
@@ -13,20 +13,88 @@
...
+    private $twig;
+
+    private $userManager;
+
+    private $adminPool;
+
     protected function setUp(): void
     {
         $this->controller = new AdminResettingController();
+        $this->container = $this->getMockBuilder(ContainerBuilder::class)->getMock();
```
OskarStark: please use `createMock()` here (and below), probably you use another (older) PHPUnit version.

This should fix the tests

kunicmarko20:
Prefer lowest is failing, on sf 2.8 `symfony/templating` is included so it is probably calling render there instead on the twig.

https://github.com/symfony/framework-bundle/blob/2.8/Controller/Controller.php#L184

tests/Controller/AdminResettingControllerTest.php
morinceline: why? `createMock` is deprecated isn't it ?

kunicmarko20: `getMock` is Deprecated IIRC

morinceline: ok :+1:

OskarStark:
Thank you

PR #1052, opened 2018-10-23
Add configurable mailer

Wharenn:
I am targeting this branch, because this is a new feature with no breaking change.

## Changelog

```markdown
### Added
- Added an optional `mailer` config option to use a custom mailer implementation
```
## Subject

By default, SonataUserBundle is using the `mailer` service to send the account password reset email and keep the email building logic in a controller private method. It makes it hard to use a custom implementation like using another mailer than the default one or updating the email creation process.

This PR adds a `Mailer` class implementing the FOS `MailerInterface` and doing the same things the `AdminResettingController` class was. This default class can be replaced by any other user service implementing the `MailerInterface` by setting corresponding service id uner the `mailer` config key:

```yaml
# config/package/sonata.yaml

sonata_user:
    mailer: custom.mailer.service.id
```

PR #1053, opened 2018-10-24
Use invokable actions

Wharenn:
I am targeting this branch, because this is backward compatible.

## Changelog

```markdown
### Changed
- `Controller\AdminResettingController` is now deprecated in favor of `Action\{CheckEmail,Request,Reset,SendEmail}Action`
- `Controller\AdminSecurityController` is now deprecated in favor of `Action\{CheckLogin,Login,Logout}Action`
```

## To do
    
- [x] Deprecated `AdminResettingController`
- [x] Deprecated `AdminSecurityController`

## Subject

As discussed in #1052, this PR deprecates `AdminResettingController` in favor of new Actions. Changes are based on the work done on this PR : https://github.com/sonata-project/SonataAdminBundle/pull/5120

This may be a first run, so comments are welcomed :)
~~I think we should also include the deprecation of `AdminSecurityController` to be consistent~~ (done)

### DIFF
--- a/src/Action/CheckEmailAction.php
+++ b/src/Action/CheckEmailAction.php
@@ -15,47 +15,63 @@ namespace Sonata\UserBundle\Action;
 
 use Sonata\AdminBundle\Admin\Pool;
 use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
-final class CheckEmailAction extends Controller
+final class CheckEmailAction
 {
+    /**
+     * @var EngineInterface
+     */
+    private $templating;
+
+    /**
+     * @var UrlGeneratorInterface
+     */
+    private $urlGenerator;
+
     /**
      * @var Pool
      */
-    protected $adminPool;
+    private $adminPool;
 
     /**
      * @var TemplateRegistryInterface
      */
-    protected $templateRegistry;
+    private $templateRegistry;
 
     /**
      * @var int
      */
-    protected $resetTtl;
+    private $resetTtl;
 
     public function __construct(
+        EngineInterface $templating,
+        UrlGeneratorInterface $urlGenerator,
         Pool $adminPool,
         TemplateRegistryInterface $templateRegistry,
         int $resetTtl
     ) {
+        $this->templating = $templating;
+        $this->urlGenerator = $urlGenerator;
         $this->adminPool = $adminPool;
         $this->templateRegistry = $templateRegistry;
         $this->resetTtl = $resetTtl;
     }
 
-    public function __invoke(Request $request)
+    public function __invoke(Request $request): Response
     {
         $username = $request->query->get('username');
 
         if (empty($username)) {
             // the user does not come from the sendEmail action
-            return new RedirectResponse($this->generateUrl('sonata_user_admin_resetting_request'));
+            return new RedirectResponse($this->urlGenerator->generate('sonata_user_admin_resetting_request'));
         }
 
-        return $this->render('@SonataUser/Admin/Security/Resetting/checkEmail.html.twig', [
+        return $this->templating->renderResponse('@SonataUser/Admin/Security/Resetting/checkEmail.html.twig', [
             'base_template' => $this->templateRegistry->getTemplate('layout'),
             'admin_pool' => $this->adminPool,
             'tokenLifetime' => ceil($this->resetTtl / 3600),
--- a/src/Action/CheckEmailAction.php
+++ b/src/Action/CheckEmailAction.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Action;
+
+use Sonata\AdminBundle\Admin\Pool;
+use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+final class CheckEmailAction extends Controller
+{
+    /**
+     * @var Pool
+     */
+    protected $adminPool;
+
+    /**
+     * @var TemplateRegistryInterface
+     */
+    protected $templateRegistry;
+
+    /**
+     * @var int
+     */
+    protected $resetTtl;
+
+    public function __construct(
+        Pool $adminPool,
+        TemplateRegistryInterface $templateRegistry,
+        int $resetTtl
+    ) {
+        $this->adminPool = $adminPool;
+        $this->templateRegistry = $templateRegistry;
+        $this->resetTtl = $resetTtl;
+    }
+
+    public function __invoke(Request $request)
+    {
+        $username = $request->query->get('username');
+
+        if (empty($username)) {
+            // the user does not come from the sendEmail action
+            return new RedirectResponse($this->generateUrl('sonata_user_admin_resetting_request'));
+        }
+
+        return $this->render('@SonataUser/Admin/Security/Resetting/checkEmail.html.twig', [
+            'base_template' => $this->templateRegistry->getTemplate('layout'),
+            'admin_pool' => $this->adminPool,
+            'tokenLifetime' => ceil($this->resetTtl / 3600),
+        ]);
+    }
+}
--- a/src/Action/CheckLoginAction.php
+++ b/src/Action/CheckLoginAction.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Action;
+
+final class CheckLoginAction
+{
+    public function __invoke(): void
+    {
+        throw new \RuntimeException('You must configure the check path to be handled by the firewall using form_login in your security firewall configuration.');
+    }
+}
--- a/src/Action/LoginAction.php
+++ b/src/Action/LoginAction.php
@@ -1,0 +1,156 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Action;
+
+use Sonata\AdminBundle\Admin\Pool;
+use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
+use Sonata\UserBundle\Model\UserInterface;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\Session;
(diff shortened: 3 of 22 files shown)


src/Action/CheckEmailAction.php
```diff
@@ -0,0 +1,64 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+final class CheckEmailAction extends Controller
+{
+    /**
+     * @var Pool
+     */
+    protected $adminPool;
```
core23: private (same for the others)

src/Action/ResetAction.php
```diff
@@ -0,0 +1,166 @@
...
+        $this->authorizationChecker = $authorizationChecker;
+        $this->router = $router;
+        $this->adminPool = $adminPool;
+        $this->templateRegistry = $templateRegistry;
+        $this->formFactory = $formFactory;
+        $this->userManager = $userManager;
+        $this->loginManager = $loginManager;
+        $this->translator = $translator;
+        $this->resetTtl = $resetTtl;
+        $this->firewallName = $firewallName;
```
core23: You need `$this->logger = new NullLogger()`

src/Controller/AdminResettingController.php
```diff
@@ -30,181 +35,47 @@ class AdminResettingController extends Controller
...
+        return $requestAction($this->getCurrentRequest());
     }
 
     /**
-     * @param Request $request
-     *
      * @return Response
      */
-    public function sendEmailAction(Request $request)
+    public function sendEmailAction()
```
core23: This is a BC break

Hanmac:
Symfony 4.2 did deprecate Controller class for AbstractController
https://symfony.com/blog/new-in-symfony-4-2-important-deprecations
maybe that can be done there too if needed?

or do you even need any base class at all?

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+     * @var Pool
+     */
+    protected $adminPool;
+
+    /**
+     * @var TemplateRegistryInterface
+     */
+    protected $templateRegistry;
+
+    public function __construct(AuthorizationCheckerInterface $authorizationChecker, RouterInterface $router, Pool $adminPool, TemplateRegistryInterface $templateRegistry)
```
greg0ire:
```suggestion
    public function __construct(
        AuthorizationCheckerInterface $authorizationChecker,
        UrlGeneratorInterface $urlGenerator,
        Pool $adminPool,
        TemplateRegistryInterface $templateRegistry
)
```

src/Action/ResetAction.php
```diff
@@ -0,0 +1,166 @@
...
+
+            try {
+                $this->loginManager->logInUser($this->firewallName, $user, $response);
+                $user->setLastLogin(new \DateTime());
+            } catch (AccountStatusException $ex) {
+                // We simply do not authenticate users which do not pass the user
+                // checker (not enabled, expired, etc.).
+                if ($this->logger) {
+                    $this->getLogger()->warning(sprintf(
+                            'Unable to login user %d after password reset',
```
greg0ire:
```suggestion
                        'Unable to login user %d after password reset',
```

src/Action/ResetAction.php
```diff
@@ -0,0 +1,166 @@
...
+                            $user->getId())
```
greg0ire:
```suggestion
                        $user->getId()
```

src/Action/ResetAction.php
```diff
@@ -0,0 +1,166 @@
...
+                    );
```
greg0ire:
```suggestion
                ), ['exception' => $ex]);
```

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+namespace Sonata\UserBundle\Action;
+
+use Sonata\AdminBundle\Admin\Pool;
+use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\RouterInterface;
```
greg0ire:
```suggestion
use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
```

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+
+final class RequestAction extends Controller
+{
+    /**
+     * @var AuthorizationCheckerInterface
+     */
+    protected $authorizationChecker;
+
+    /**
+     * @var RouterInterface
```
greg0ire:
```suggestion
     * @var UrlGeneratorInterface
```

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    protected $router;
```
greg0ire:
```suggestion
    private $urlGenerator;
```

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @var TemplateRegistryInterface
+     */
+    protected $templateRegistry;
+
+    public function __construct(AuthorizationCheckerInterface $authorizationChecker, RouterInterface $router, Pool $adminPool, TemplateRegistryInterface $templateRegistry)
+    {
+        $this->authorizationChecker = $authorizationChecker;
+        $this->router = $router;
```
greg0ire:
```suggestion
        $this->urlGenerator = $urlGenerator;
```

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->adminPool = $adminPool;
+        $this->templateRegistry = $templateRegistry;
+    }
+
+    public function __invoke(Request $request)
```
greg0ire:
```suggestion
    public function __invoke(Request $request): Response
```

src/Action/RequestAction.php
```diff
@@ -0,0 +1,65 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Action;
+
+use Sonata\AdminBundle\Admin\Pool;
+use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
```
greg0ire:
```suggestion
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
```

src/Action/ResetAction.php
```diff
@@ -0,0 +1,166 @@
...
+        $this->templateRegistry = $templateRegistry;
+        $this->formFactory = $formFactory;
+        $this->userManager = $userManager;
+        $this->loginManager = $loginManager;
+        $this->translator = $translator;
+        $this->resetTtl = $resetTtl;
+        $this->firewallName = $firewallName;
+    }
+
+    public function __invoke(Request $request, $token)
```
greg0ire:
```suggestion
    public function __invoke(Request $request, $token): Response
```

src/Action/ResetAction.php
```diff
@@ -0,0 +1,166 @@
...
+
+use FOS\UserBundle\Form\Factory\FactoryInterface;
+use FOS\UserBundle\Model\UserManagerInterface;
+use FOS\UserBundle\Security\LoginManagerInterface;
+use Psr\Log\LoggerAwareTrait;
+use Sonata\AdminBundle\Admin\Pool;
+use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
```
greg0ire:
```suggestion
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
```

src/Action/CheckEmailAction.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Sonata\UserBundle\Action;
+
+use Sonata\AdminBundle\Admin\Pool;
+use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+final class CheckEmailAction extends Controller
```
greg0ire: Bonus points if you avoid extending `Controller` by injecting the templating engine

Hanmac:
@greg0ire your opinion to the Controller depreciation or do we not care yet?

OskarStark:
I think he means there is no need to extend anything

greg0ire:
> I think he means there is no need to extend anything

Yes, but as I said, "for bonus points", meaning I will not block this already great PR with that request.

Wharenn:
I made an update by adding the actions of the `AdminSecurityController` and also including removal of the extends (clean actions yeah!). I think the scope of this PR is now complete. 

@greg0ire @core23, if there is any new changes you see, I will try to update asap :)

src/Action/RequestAction.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function __invoke(Request $request): Response
+    {
+        if ($this->authorizationChecker->isGranted('IS_AUTHENTICATED_FULLY')) {
+            return new RedirectResponse($this->urlGenerator->generate('sonata_admin_dashboard'));
+        }
+
+        return $this->templating->renderResponse('@SonataUser/Admin/Security/Resetting/request.html.twig', [
+            'base_template' => $this->templateRegistry->getTemplate('layout'),
+            'admin_pool' => $this->adminPool,
```
core23: Not part of this PR, but can we make a global twig variable for `base_template` and `admin_pool`, as every admin template need this

src/Action/ResetAction.php
```diff
@@ -0,0 +1,182 @@
...
+     * @var int
+     */
+    private $resetTtl;
+
+    /**
+     * @var string
+     */
+    private $firewallName;
+
+    public function __construct(
```
core23: OMG! Do we really need that much deps? oO

core23:
I like the PR 👍  but should we pick the chance to reduce the dependencies for the classes first? Later that would be a BC break.

src/Action/LoginAction.php
```diff
@@ -0,0 +1,149 @@
...
+
+            return new RedirectResponse($url);
+        }
+
+        $csrfToken = null;
+        if ($this->csrfTokenManager) {
+            $csrfToken = $this->csrfTokenManager->getToken('authenticate')->getValue();
+        }
+
+        return $this->templating->renderResponse('@SonataUser/Admin/Security/login.html.twig', [
```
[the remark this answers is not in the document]
Wharenn: It looks like the `templating` service is defined by FrameworkBundle which is a requirement. So I think we always fell to this definition in case of rendering:
https://github.com/symfony/framework-bundle/blob/master/Resources/config/templating_php.xml
https://github.com/symfony/framework-bundle/blob/4.1/DependencyInjection/FrameworkExtension.php#L823

src/Action/CheckLoginAction.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Action;
+
+final class CheckLoginAction
```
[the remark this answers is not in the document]
Wharenn: We need the routing of these actions for the login/logout process. What do you propose? A single catch action throwing an exception?

Wharenn:
Do you see any clue to reduce the amount of dependencies? I do not see any optional dependency here. Dependencies listing is the "downside" of container-less controllers/actions as we're not relying anymore on the underlying container `get()` method for common actions like translations, rendering or flash messages.

core23:
> Do you see any clue to reduce the amount of dependencies? 

Maybe we can use Events (like https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Controller/ChangePasswordController.php) for some stuff, e.g. Flash Messages and Mailing

> I do not see any optional dependency here. Dependencies listing is the "downside" of container-less controllers/actions as we're not relying anymore on the underlying container `get()` method for common actions like translations, rendering or flash messages.

I know that the old Controllers had way more intransparent dependencies, the new action style is a huge improvement and made all dependencies transparent 👍

Wharenn:
This approach is interesting. As you proposed to remove dependencies on the pool and template registry, it could be really cool to get rid of the session/flash with events. Maybe we could do this in a next PR. I must confess this has been a long run since the #1052 initial proposal 😅

Hanmac:
@core23 interesting, i wonder because UserBundle already require FOSUser, so in theory we can use its events too?

greg0ire:
Keep in mind that if this is done in a subsequent PR, it will be a BC-break unless we do not create a new release. Otherwise we will have to keep the extra constructor arguments and deprecate them.

OskarStark:
@core23 can you approve and if so, please merge :)

PR #1054, opened 2018-11-10
DevKit updates for 4.x branch

SonataCI:


.travis.yml
```diff
@@ -37,35 +38,36 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: '7.2'
+    - php: '7.3'
```
OskarStark: Damn we need a new co fog on dev kit for the default php version, because this is still 7.2 and not 7.3, it should only be added to the Matrix an be allowed to fail ...

PR #1057, opened 2018-11-16
updated bootstrap.php file

OskarStark:
Refs https://github.com/sonata-project/dev-kit/issues/436

see my comment

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /*
  * This file is part of the Sonata Project package.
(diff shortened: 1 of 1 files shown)


tests/bootstrap.php
```diff
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
```
OskarStark: if we want to manage these files via dev-kit, we either need to remove this and support our lowest PHP version (organization wide) OR need to add it as a template to dev-kit, parse it and do some if-clauses in the template, depending on the "default project php version"

PR #1067, opened 2019-01-02
[Docs] small changes

OskarStark:


docs/reference/two_step_validation.rst
```diff
@@ -21,15 +21,17 @@ So if your login and password are compromised then the hacker must also hold you
 Installation
 ------------
 
-Add the following line to the ``composer.json`` file::
+Add the following line to the ``composer.json`` file:
 
-    "sonata-project/google-authenticator": "dev-master"
+.. code-block:: bash
+
+    composer require sonata-project/google-authenticator
```
OskarStark: this is important, because we now have a stable version!

kunicmarko20: The text (Add the following line to the ``composer.json`` file:)  doesn't make sense now, please update that too

OskarStark: done, thx!

PR #1095, opened 2019-05-29
Allow using external model classes

### LINKED ISSUE #967: [4.0] way to disable checkManagerType or allow FOSUser as parent?

javer:
## Subject

Allow using external model classes.

See #980 for the previous conversation.

I am targeting this branch, because this is BC since commit 505de3e.

Closes #967

## Changelog

```markdown
### Fixed
- Fixed a bug with inability to use external model classes
```

### DIFF
--- a/src/DependencyInjection/SonataUserExtension.php
+++ b/src/DependencyInjection/SonataUserExtension.php
@@ -98,7 +98,7 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
             $loader->load('security_acl.xml');
         }
 
-        $this->checkManagerTypeToModelTypeMapping($config);
+        $this->checkManagerTypeToModelTypesMapping($config);
 
         $this->registerDoctrineMapping($config);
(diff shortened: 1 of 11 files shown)


javer:
@OskarStark Could you please look at proposed fix? I have also fixed all PHPUnit deprecations as well.

javer:
Any update on this?

OskarStark:
Sorry I missed your ping

PR #1100, opened 2019-06-14
Replace usages of deprecated `spaceless` Twig tag

phansys:
## Subject

Replace usages of deprecated `spaceless` Twig tag.

I am targeting this branch, because the changes are considered BC.

## Changelog

```markdown
### Changed
- Add missing dependency against "twig/twig";
- Changed usages of `{% spaceless %}` tag, which is deprecated as of Twig 1.38 with `{% apply spaceless %}` filter;
```

Follow up of sonata-project/SonataAdminBundle#5576.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
         "symfony/options-resolver": "^2.8 || ^3.2 || ^4.0",
         "symfony/security-acl": "^2.8 || ^3.0",
         "symfony/security-core": "^2.8 || ^3.2 || ^4.0",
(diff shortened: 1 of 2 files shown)


kunicmarko20:
Should we do this on stable? forcing people to use the newest twig?

phansys:
> Should we do this on stable? forcing people to use the newest twig?

Please, see https://github.com/sonata-project/SonataAdminBundle/issues/5573#issuecomment-499493869.

PR #1088, opened 2019-04-11
Allow usage of JMS SerializerBundle 3.0

supersmile2009:
## Subject

I am targeting this branch, because it's a backwards compatible version constraint modification.
None of the BC breaks of a new major version of JMS SerializerBundle seem to affect common use cases found in SonataUserBundle. Therefore no special code modifications are required.

Closes #1082

## Changelog
```markdown
### Added
- Added compatibility with jms/serializer-bundle:^3.0 and jms/serializer:^2.0
```

PR #1103, opened 2019-06-22
Allow "jms/serializer:^3.0"

phansys:
## Subject

Allow "jms/serializer:^3.0"

I am targeting this branch, because these changes are BC.

## Changelog

```markdown
### Removed
- Dependency conflict against "jms/serializer:^3.0".
```

I don't know why this is added as conflicting version at #1088:
https://github.com/sonata-project/SonataUserBundle/blob/9082b5c06ddd0fa3f1ed39582a92f53bfcbc768a/composer.json#L46

but I'll try to add some tests to ensure there is no BC with that version.
See https://github.com/schmittjoh/serializer/blob/master/UPGRADING.md#from-2x-to-300.

## To do
    
- [x] Add tests;
- [x] Fix tests for serialized date formats.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.1 || >=3.0",
-        "jms/serializer": "<0.13 || >=3.0",
+        "jms/serializer": "<1.3 || >=4.0",
         "nelmio/api-doc-bundle": "<2.4",
         "sonata-project/block-bundle": "<3.11",
         "sonata-project/doctrine-orm-admin-bundle": "<3.0",
--- a/tests/Serialization/SerializationTest.php
+++ b/tests/Serialization/SerializationTest.php
@@ -1,0 +1,250 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Tests\Serialization;
+
+use JMS\Serializer\Handler\DateHandler;
+use JMS\Serializer\Handler\HandlerRegistryInterface;
+use JMS\Serializer\SerializationContext;
+use JMS\Serializer\SerializerBuilder;
+use PHPUnit\Framework\TestCase;
+use Sonata\UserBundle\Tests\Entity\User;
+
+/**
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -44,7 +44,7 @@
     },
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.1 || >=3.0",
-        "jms/serializer": "<0.13 || >=3.0",
+        "jms/serializer": "<1.3 || >=4.0",
```
kunicmarko20: title says "Allow jms serializer ^3.0" but this change makes it impossible to install serializer 3.0 https://jubianchi.github.io/semver-check/#/%3C1.3%20||%20%3E%3D4.0/3.0.0 or am I missing something?

phansys: This constraint is at the "conflict" section, it disallows versions lower to `1.3.0` and higher or equal than `4.0.0`.

kunicmarko20: ah, sry, didn't see it was in conflict

greg0ire:
:thinking: the ` "friendsofsymfony/rest-bundle": "<2.1 || >=3.0", ` line was not touched in the PR you linked…

phansys:
Sorry, I have linked the previous line. Description updated.

greg0ire:
Oh ok right makes sense. This line was added so as to let that dependency be optional, while still telling the user what version they should use if they install it.

phansys:
>Oh ok right makes sense. This line was added so as to let that dependency be optional, while still telling the user what version they should use if they install it.

That's fine. What I don't understand is why 3.x branch was restricted. I didn't see any feature provided by this bundle that could stop working with that.

greg0ire:
I suppose it did not exist at the time, and we didn't know what would break when it would, maybe?

phansys:
You're right, #1088 was created at 2019-04-13, while [3.0.0](https://github.com/schmittjoh/serializer/releases/tag/3.0.0) version was released at 2019-04-23 :+1:

phansys:
The CS rules are disallowing the usage of `TestCase::assertEquals()`, but I'd like to avoid comparing 2 arrays strictly with `TestCase::assertSame()`, since in this case the order of their values could change without breaking anything. Do you know if there is another way to perform the same check without this limitation?

greg0ire:
I think you could sort both arrays before comparing them strictly?

phansys:
I was expecting some fancier solution :smile:
For some reason, `array_multisort()` is not generating the same results. I'll try to push again when the tests will be passing.

phansys:
Well, `ksort()` seems to work fine with these structures. Changes pushed.

greg0ire:
Thanks @phansys !

PR #1104, opened 2019-06-27
Do not load default mailer service when using custom mailer service

TomAdam:
## Subject

The default mailer has a hard dependency on the Symfony "mailer" service which is not available in all use cases. This change prevents loading the default SonataUserBundle mailer service when a custom mailer service is provided.

I am targeting this branch, because this PR is a patch for an issue. I consider the default SonataUser mailer service to be internal and it is unlikely that external code depends on it when an alternative mailer is set. Under that assumption this is not a BC break.

## Changelog

```markdown
### Changed
- Do not register `sonata.user.mailer.default` service when a custom mailer service is defined
```

### DIFF
--- a/src/DependencyInjection/SonataUserExtension.php
+++ b/src/DependencyInjection/SonataUserExtension.php
@@ -81,7 +81,10 @@ class SonataUserExtension extends Extension implements PrependExtensionInterface
         $loader->load('twig.xml');
         $loader->load('command.xml');
         $loader->load('actions.xml');
-        $loader->load('mailer.xml');
+
+        if ('sonata.user.mailer.default' === $config['mailer']) {
+            $loader->load('mailer.xml');
+        }
 
         if ('orm' === $config['manager_type'] && isset(
             $bundles['FOSRestBundle'],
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/SonataUserExtension.php
```diff
@@ -81,7 +81,10 @@ class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator')) {
         $loader->load('twig.xml');
         $loader->load('command.xml');
         $loader->load('actions.xml');
-        $loader->load('mailer.xml');
+
+        if ('sonata.user.mailer.default' === $config['mailer']) {
```
phansys: Even if it shouldn't be a valid use case, I think we should check against the configured class too, since the service definition could be overridden by the user if I'm not wrong.

TomAdam: As an alternative we could make the service non-public. It would get optimised out of the container if it was unused.

phansys: You're right, I guess. But that would a BC break AFAIK.

TomAdam: > I think we should check against the configured class too

It is possible that someone has injected the default mailer into something else, as well as replacing the service definition. I'm thinking this is going to be a BC break however we look at it. Should I rebase on master and make the service non-public rather than not loading the service definition? Feels cleaner to me.

PR #1105, opened 2019-06-28
Removed deprecation

lukepass:
## Subject

Changed deprecated class and removed CoreBundle class.

## Changelog

```markdown
### Fixed
- Changed deprecated class.
```


phansys:
@lukepass, thank you so much for the PR.
Could you please respect the PR template? I'm suspecting these changes are not BC.

greg0ire:
>  Could you please respect the PR template? I'm suspecting these changes are not BC.

They should be, IIRC some of these names are just aliases for the new names.
However, their is a tiny CS issue that can be fixed with `make cs-fix` and there are probably constraints to bump so that we are sure the new classes are available

phansys:
By instance, `Sonata\DatagridBundle\Pager\PageableInterface` doesn't extend `Sonata\CoreBundle\Model\PageableManagerInterface`, so AFAIK `Sonata\UserBundle\Model\GroupManagerInterface` will not be an instance of  `Sonata\DatagridBundle\Pager\PageableInterface` anymore.

See https://github.com/sonata-project/SonataDatagridBundle/blob/855fa7da7b0515e9f424f6ce7ca1d3a13668a154/src/Pager/PageableInterface.php#L16

greg0ire:
Fixed it :)

phansys:
I don't understand how is it fixed. I've checked out this PR locally and I still see the scenario I've shared before:
>Sonata\DatagridBundle\Pager\PageableInterface doesn't extend Sonata\CoreBundle\Model\PageableManagerInterface, so AFAIK Sonata\UserBundle\Model\GroupManagerInterface will not be an instance of Sonata\DatagridBundle\Pager\PageableInterface anymore.

greg0ire:
Oh sorry, I didn't notice your earlier message. As you can see, one is just an alias for the other: https://github.com/sonata-project/SonataCoreBundle/blob/a5d2ce61de980a8eb53613e3c49a24c4d6794f01/src/CoreBundle/Model/PageableManagerInterface.php#L24-L27

greg0ire:
:thinking: but it has really worked well since 3.14.1, so let me bump the core version constraint too.

phansys:
Sorry, but I really don't see `PageableInterface` as an alias or subclass of `PageableManagerInterface`. Maybe I'm missing something...

greg0ire:
Oh right, my bad, I didn't see the `Manager` part…

greg0ire:
I think that message is wrong: https://github.com/sonata-project/SonataCoreBundle/blob/a5d2ce61de980a8eb53613e3c49a24c4d6794f01/src/CoreBundle/Model/PageableManagerInterface.php#L19

greg0ire:
Oh but the https://github.com/sonata-project/sonata-doctrine-extensions/blob/1.x/src/Model/PageableManagerInterface.php#L27 is deprecated in favor of the `PageableInterface`… so things might actually be right.

greg0ire:
And there is no alias between them as you noticed so indeed, BC break… not sure what to do… @core23 , help? :sweat_smile:

greg0ire:
I pushed a new commit where 2 interfaces are implemented… that way people can migrate to the best one, but we should probably work on another class_alias to make it BC

core23:
I think we should not spend to much time on class_aliases and other ways for a soft migration. Realising new major version of our bundles is quite hard because we have too many hard couplings. 

Removing the CoreBundle without any pain is not possible for our small team at the moment.

greg0ire:
Let's merge as is then?

greg0ire:
Oh right. There is an intermediary step though, that would be to follow the first deprecation, and that wouldn';t be a BC-break (using `Sonata\DatagridBundle\Pager\PagerInterface`).

core23:
@lukepass Can you rebase this against the master branch please

lukepass:
Yes, but I am not really sure how to do it?

core23:
> Yes, but I am not really sure how to do it?

Sorry for the delay. 


Given you have checked out your fork as `origin` and the sonata repository as `upstream`:

```
git fetch
git rebase upstream/4.x
git push -f
```

PR #1110, opened 2019-08-07
Use "twig" service instead of deprecated "templating" where possible

phansys:
## Subject

Use "twig" service instead of deprecated "templating" where possible

## Changelog

```markdown
### Changed
- Replaced usages of deprecated "templating" service with "twig" where possible.
```

### DIFF
--- a/src/GoogleAuthenticator/RequestListener.php
+++ b/src/GoogleAuthenticator/RequestListener.php
@@ -14,10 +14,12 @@ declare(strict_types=1);
 namespace Sonata\UserBundle\GoogleAuthenticator;
 
 use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\HttpKernel;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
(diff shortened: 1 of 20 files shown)


core23:
Should we do this on the stable branch?

src/GoogleAuthenticator/RequestListener.php
```diff
@@ -32,15 +34,53 @@ class RequestListener
...
+     * @var Environment
+     */
+    private $twig;
+
+    /**
+     * NEXT_MAJOR: Remove `$templating` argument and make `$twig` argument mandatory
+     *
+     * @param EngineInterface|Environment $templating
+     */
+    public function __construct(Helper $helper, TokenStorageInterface $tokenStorage, object $templating = null, Environment $twig = null)
```
phansys: I have also updated this class in order to use "twig" service if possible, respecting the BC promise. This is the only non final class which requires this logic.
I guess I'll send another PR to close the API with `@final` annotations covering this class too.

tests/Action/ResetActionTest.php
```diff
@@ -31,56 +30,57 @@
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Symfony\Component\Translation\TranslatorInterface;
+use Twig\Environment;
 
 class ResetActionTest extends TestCase
 {
     /**
-     * @var EngineInterface|\PHPUnit_Framework_MockObject_MockObject
+     * @var Environment|\PHPUnit\Framework\MockObject\MockObject
```
core23: Please do not use FCQNs here

PR #1113, opened 2019-09-19
Get switch user role from current firewall config

sad270:
## Subject

In UserAdmin we use 'ROLE_ALLOWED_TO_SWITCH' role to show 'impersonating' action in User list
This role is the default name for switch user role, and can be change in firewall config.

```yaml
# app/config/security.yml
security:
    firewalls:
        firewall_name:
            switch_user:
                role: ANOTHER_ROLE_ALLOWED_TO_SWITCH
```

If we change role name in firewall config. The new role is not used in UserAdmin.

I am targeting this branch, because this is a patch with BC.

## Changelog

```markdown
### Fixed
- Switch user role from current firewall config is now used instead of 'ROLE_ALLOWED_TO_SWITCH' to show impersonating action in User list
```

### DIFF
--- a/src/Admin/Model/UserAdmin.php
+++ b/src/Admin/Model/UserAdmin.php
@@ -22,6 +22,7 @@ use Sonata\AdminBundle\Form\Type\ModelType;
 use Sonata\AdminBundle\Show\ShowMapper;
 use Sonata\CoreBundle\Form\Type\DatePickerType;
 use Sonata\UserBundle\Form\Type\SecurityRolesType;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\LocaleType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
@@ -99,7 +100,7 @@ class UserAdmin extends AbstractAdmin
             ->add('createdAt')
         ;
 
-        if ($this->isGranted('ROLE_ALLOWED_TO_SWITCH')) {
+        if ($this->canImpersonate()) {
             $listMapper
                 ->add('impersonating', 'string', ['template' => '@SonataUser/Admin/Field/impersonating.html.twig'])
             ;
@@ -250,5 +251,20 @@ class UserAdmin extends AbstractAdmin
                 ->end()
             ->end()
         ;
+    }
+
+    private function canImpersonate(): bool
+    {
+        if (!$this->getConfigurationPool()->getContainer()->getParameter('sonata.user.impersonating')) {
+            return false;
+        }
+        $role = 'ROLE_ALLOWED_TO_SWITCH';
+        $firewallMap = $this->getConfigurationPool()->getContainer()->get('security.firewall.map', ContainerInterface::NULL_ON_INVALID_REFERENCE);
+        $firewallConfig = $firewallMap ? $firewallMap->getFirewallConfig($this->getRequest()) : null;
+        if ($firewallConfig && null !== $switchUserConfig = $firewallConfig->getSwitchUser()) {
+            $role = $switchUserConfig['role'];
+        }
+
+        return $this->isGranted($role);
     }
 }


src/Admin/Model/UserAdmin.php
```diff
@@ -251,4 +251,16 @@ protected function configureFormFields(FormMapper $formMapper): void
             ->end()
         ;
     }
+
+    protected function isAllowedToSwitch(): bool
```
core23: No need to make this `protected`

core23 (changes requested):

src/Admin/Model/UserAdmin.php
sad270: `isAllowedToSwitch` method will only be used in `UserAdmin` class.
No need to be `public`.
So i make it `private` ?

core23: You should always use the lowest visibility, if possible

src/Admin/Model/UserAdmin.php
```diff
@@ -251,4 +251,16 @@ protected function configureFormFields(FormMapper $formMapper): void
             ->end()
         ;
     }
+
+    private function isAllowedToSwitch(): bool
+    {
+        $role = 'ROLE_ALLOWED_TO_SWITCH';
+        $firewallMap = $this->getConfigurationPool()->getContainer()->get('security.firewall.map');
```
phansys: Since this configuration is [provided by "symfony/security-bundle"](https://github.com/symfony/security-bundle/blob/97ba8648e718999793e79ab4d1f1582c8d19be9d/DependencyInjection/MainConfiguration.php#L264-L275), I think we should check for the availability of this 
 configuration before, or add the requirement explicitly.

phansys (changes requested):

OskarStark:
Can you show where it’s hardcored? I am not sure we should add such feature, as it’s less common IMO

src/Admin/Model/UserAdmin.php
sad270: I add `->get('security.firewall.map', ContainerInterface::NULL_ON_INVALID_REFERENCE)` and check if firewall.map is not null.
Is it good ?

sad270:
> Can you show where it’s hardcored? I am not sure we should add such feature, as it’s less common IMO
The `ROLE_ALLOWED_TO_SWITCH` is only hardcored here https://github.com/sonata-project/SonataUserBundle/pull/1113/files#diff-639f5f2668e6c55cd2485d04aae8a378L102

I also check if impersonating is configured in sonata to not show an empty column `$this->getConfigurationPool()->getContainer()->getParameter('sonata.user.impersonating')`

src/Admin/Model/UserAdmin.php
```diff
@@ -99,7 +100,9 @@ protected function configureListFields(ListMapper $listMapper): void
             ->add('createdAt')
         ;
 
-        if ($this->isGranted('ROLE_ALLOWED_TO_SWITCH')) {
+        if ($this->getConfigurationPool()->getContainer()->getParameter('sonata.user.impersonating')
```
core23: Can you move this to `isAllowedToSwitch`

core23 (changes requested):

src/Admin/Model/UserAdmin.php
sad270: Ok should i rename `isAllowedToSwitch` to `canImpersonating`
Or rename it `addImpersonating` rewrite it like this :
```php
    /**
     * @param ListMapper $listMapper
     */
    private function addImpersonating(ListMapper $listMapper): void
    {
        if (!$this->getConfigurationPool()->getContainer()->getParameter('sonata.user.impersonating')) {
            return;
        }
        $role = 'ROLE_ALLOWED_TO_SWITCH';
        $firewallMap = $this->getConfigurationPool()->getContainer()->get('security.firewall.map', ContainerInterface::NULL_ON_INVALID_REFERENCE);
        $firewallConfig = $firewallMap ? $firewallMap->getFirewallConfig($this->getRequest()) : null;
        if ($firewallConfig && null !== $switchUserConfig = $firewallConfig->getSwitchUser()) {
            $role = $switchUserConfig['role'];
        }

        if ($this->isGranted($role)) {
            $listMapper
                ->add('impersonating', 'string', ['template' => '@SonataUser/Admin/Field/impersonating.html.twig'])
            ;
        }
    }
```

core23: :+1: canImpersonating

src/Admin/Model/UserAdmin.php
```diff
@@ -251,4 +252,19 @@ protected function configureFormFields(FormMapper $formMapper): void
...
+        }
+        $role = 'ROLE_ALLOWED_TO_SWITCH';
+        $firewallMap = $this->getConfigurationPool()->getContainer()->get('security.firewall.map', ContainerInterface::NULL_ON_INVALID_REFERENCE);
+        $firewallConfig = $firewallMap ? $firewallMap->getFirewallConfig($this->getRequest()) : null;
+        if ($firewallConfig && null !== $switchUserConfig = $firewallConfig->getSwitchUser()) {
+            $role = $switchUserConfig['role'];
+        }
+
+        return $this->isGranted($role);
+    }
```
greg0ire: We should really refrain from accessing the container like this IMO. Besides, the role is not going to change at runtime. IMO this should be done with configuration, and the configuration node could be changed automatically if not set, by reading the configuration of the security bundle, like so: https://symfony.com/doc/4.1/bundles/prepend_extension.html

core23 (approved):

PR #1117, opened 2019-10-04
Fix non-existent method call in ResetAction

ofetisov:
## Fix non-existent "getLogger" method call in ResetAction

In ResetAction class there is a call of non-existent method "getLogger". This issue will reproduce if you'll reset a password for disabled user. 

I am targeting this branch, because this is a small bug fix.

```markdown
### Fixed
 - Fixed invocation of non-existent "getLogger" method. Changed to access property.
```


greg0ire:
Thanks @ofetisov !

PR #1118, opened 2019-10-06
Allow using all types of TokenInterface

dnna:
## Subject

The current implementation of the Google Authenticator Helper only allows the use of `UsernamePasswordToken` (e.g. `form_login`). This prevents combining 2FA with bundles such as https://github.com/hwi/HWIOAuthBundle .

This PR aims to solve this, by allowing all types of TokenInterface to be used.

I am targeting 4.x, because I'm not sure whether this change is considered backward compatible or not, so I'm targetting the most conservative branch.

## Changelog

```markdown

### Changed
- 2FA authentication now supports all types of tokens instead of only UsernamePasswordToken
```

### DIFF
--- a/src/GoogleAuthenticator/Helper.php
+++ b/src/GoogleAuthenticator/Helper.php
@@ -94,12 +94,12 @@ class Helper
      */
     public function getSessionKey(TokenInterface $token)
     {
-        if (method_exists($token,'getProviderKey')) {
+        if (method_exists($token, 'getProviderKey')) {
             $providerKey = $token->getProviderKey();
         } else {
             $providerKey = '';
         }
-        
+
         return sprintf('sonata_user_google_authenticator_%s_%s', $providerKey, $token->getUsername());
     }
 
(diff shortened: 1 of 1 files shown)


src/GoogleAuthenticator/Helper.php
```diff
@@ -92,9 +92,15 @@ public function generateSecret()
     /**
      * @return string
      */
-    public function getSessionKey(UsernamePasswordToken $token)
+    public function getSessionKey(TokenInterface $token)
     {
-        return sprintf('sonata_user_google_authenticator_%s_%s', $token->getProviderKey(), $token->getUsername());
+        if (method_exists($token, 'getProviderKey')) {
```
core23: Please check for the instance instead of method existence

dnna: That narrows the scope (couldn't there be other token types than UsernamePasswordToken that contain getProviderKey?), but I'm OK with it. I have made the change but FlintCI is now failing on some unrelated files.

dnna:
Can anyone help with what flintci/pr is complaining about? I think this issue is still relevant.

PR #1130, opened 2019-12-10
Remove CoreBundle dependency

franmomu:
## Subject

I'm not sure if `UserSerializerHandler` is needed.

I am targeting this branch, because these changes are not BC.

## Changelog

```markdown
### Removed
- Drop CoreBundle dependency
```


src/Serializer/UserSerializerHandler.php
```diff
@@ -13,18 +13,81 @@
...
+
+        return $methods;
+    }
+
+    /**
+     * Serialize data object to id.
+     *
+     * @param object $data
+     *
+     * @return int|null
```
phansys: Couldn't the object use a string for its id? (like UUID by instance).

franmomu: That's why I thought if this class is used or not. I haven't found any reference to `serialized-name` or `'sonata_user_user_id'` like the [doc says](https://github.com/sonata-project/SonataCoreBundle/pull/41/files#diff-4a1e47a8a97d01d2f85664bbce2ddf2dR24).

greg0ire:
Are there no ways to avoid the BC-breaks and contribute this to the stable branch instead?

core23:
> Are there no ways to avoid the BC-breaks and contribute this to the stable branch instead?

IMHO, we *could* this on the stable branch with some hacks and proxy classes, but it would be very time-consuming and we should spend our energy on the major releases.

This PR will change some base classes, so we have technically a new API on each class.

greg0ire:
OK, let's merge like this then.

greg0ire:
Thanks @franmomu !

PR #1133, opened 2019-12-18
Tests clean up

franmomu:
I was going to fix the `EntityManagerMockFactoryTrait` and while I was at it I did some clean up adding type declaration and so on.


core23:
Thanks @franmomu

PR #1182, opened 2020-06-12
Manual merge

greg0ire:


composer.json
```diff
@@ -27,7 +27,7 @@
         "sonata-project/datagrid-bundle": "^2.4.0",
         "sonata-project/doctrine-extensions": "^1.5.1",
         "sonata-project/easy-extends-bundle": "^2.2",
-        "sonata-project/form-extensions": "^1.1",
+        "sonata-project/form-extensions": "^0.1 || ^1.4",
```
jordisala1991, line 30: I think we should leave only 1.4

Issue #1174, opened 2020-05-26
`\Sonata\UserBundle\Action\ResetAction` is relying on concrete implementation of Session


oleg-andreyev:
#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.49.1 3.66.0 The missing Symfony Admin Generator
sonata-project/block-bundle              3.18.5 4.2.0  Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.0.1  Cache library
sonata-project/core-bundle               3.17.2 3.19.0 Symfony SonataCoreBundle (abandoned)
sonata-project/datagrid-bundle           2.5.0  3.2.0  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.6.0  1.6.0  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.13.0 3.17.1 Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.5.0  2.5.0  Symfony SonataEasyExtendsBundle
sonata-project/exporter                  2.2.0  2.2.0  Lightweight Exporter library
sonata-project/jquery-bundle             1.10.2 1.10.2 Symfony SonatajQueryBundle
Package sonata-project/jquery-bundle is abandoned, you should avoid using it. No replacement was suggested.
sonata-project/user-bundle               4.4.0  4.5.3  Symfony SonataUserBundle
```

#### Symfony packages

```
$ composer show --latest 'symfony/*'
symfony/asset                 v3.4.40 v5.0.8  Symfony Asset Component
symfony/assetic-bundle        v2.8.2  v2.8.2  Integrates Assetic into Symfony2
Package symfony/assetic-bundle is abandoned, you should avoid using it. Use symfony/webpack-encore-pack instead.
symfony/browser-kit           v3.4.40 v5.0.8  Symfony BrowserKit Component
symfony/cache                 v3.4.40 v5.0.8  Symfony Cache component with PSR-6, PSR-16, and tags
symfony/class-loader          v3.4.40 v3.4.40 Symfony ClassLoader Component
symfony/config                v3.4.40 v5.0.8  Symfony Config Component
symfony/console               v3.4.40 v5.0.8  Symfony Console Component
symfony/css-selector          v3.4.40 v5.0.8  Symfony CssSelector Component
[...]

PR #1184, opened 2020-06-20
Revert of previous commit. Actual code relays on Session class.

kubec:
## Actual code relays on Session class

Actual code relays on `Session` class. Not `SessionInterface` class. Problem is in the method `getFlashBag()` of `Session` class. This method is widely used in code in many classes. If we want relay on  `SessionInterface` class, we need to solve the problem with using `getFlashBag()`.

It is revert of changes from the #1174 

## Changelog
Actual code still relays on `Session` class, not `SessionInterface` class. 

```markdown
### Changed
- Reverting use of Session class instead of SessionInterface class
```

### DIFF
--- a/src/Action/ResetAction.php
+++ b/src/Action/ResetAction.php
@@ -23,7 +23,7 @@ use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
@@ -81,7 +81,7 @@ final class ResetAction
     private $translator;
 
     /**
-     * @var SessionInterface
+     * @var Session
      */
     private $session;
 
@@ -105,7 +105,7 @@ final class ResetAction
         UserManagerInterface $userManager,
         LoginManagerInterface $loginManager,
         TranslatorInterface $translator,
-        SessionInterface $session,
+        Session $session,
         int $resetTtl,
         string $firewallName
     ) {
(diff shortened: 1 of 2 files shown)


greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

greg0ire:
@oleg-andreyev please review

oleg-andreyev:
The problem with relaying on concrete implementation is that user may want to decorate Session. As an example with logging (it’s my case). 

Probably this case should solved with PR to Symfony repository. 

We need to have new interface in FrameworkBundle which will define flashbag methods and extend SessionInterface from HttpFoundation, but doing so could be a massive BC break.

Since currently Session class is the only implementation of SessionInterface I think we should leave type in controller and have concrete Session in tests - it’s a bit fragile but I don’t see another solution.

wbloszyk:
Mybe it is possible to keep `SessionInterface`. 
```php
method_exists(self, 'getFlashBag') ? 
        $this->session = $session :
        $this->session = new FlashBagSessionDecorator($session); 
```
It is possible to write working decorator?
WDYT? @oleg-andreyev

kubec:
@oleg-andreyev I understand your problem but it is too fragile for me. This problem is in the class `LoginAction` too. Behavior should be consistent.

@wbloszyk It could be the corret way.

phansys:
For the PR description, please note that the reverted PR is #1175.

oleg-andreyev:
Looks like I've found a solution for both problems, we can inject `session.flash_bag` which is an instance of `FlashBagInterface` and is build using "session" factory

```
<service id="session.flash_bag" class="Symfony\Component\HttpFoundation\Session\Flash\FlashBag">
  <factory service="session" method="getFlashBag"/>
</service>
<service id="Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface" alias="session.flash_bag" />
```

src/Action/ResetAction.php
```diff
@@ -105,7 +105,7 @@ public function __construct(
         UserManagerInterface $userManager,
         LoginManagerInterface $loginManager,
         TranslatorInterface $translator,
-        SessionInterface $session,
+        Session $session,
```
oleg-andreyev, line 108: Please take a look at https://github.com/sonata-project/SonataUserBundle/pull/1184#issuecomment-647620411

oleg-andreyev, line 108: Could you please take a look at https://github.com/sonata-project/SonataUserBundle/pull/1184#issuecomment-647620411

jordisala1991:
What is the status of this?

PR #1185, opened 2020-06-22
Allow "friendsofsymfony/rest-bundle:^3.0"

phansys:
## Subject

Allow "friendsofsymfony/rest-bundle:^3.0".

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Added
- Added support for "friendsofsymfony/rest-bundle:^3.0";
- Added public alias `Sonata\UserBundle\Controller\Api\UserController` for `sonata.user.controller.api.user` service;
- Added public alias `Sonata\UserBundle\Controller\Api\GroupController` for `sonata.user.controller.api.group` service.
### Removed
- Removed support for symfony/*:<4.4;
- Removed support for deprecated "rest" routing type.
```
Sonata\UserBundle\Controller\Api\GroupController

## To do

- [x] Update the tests;
- [x] Update the docs;
- [x] Update the routing in order to use other route type than "rest".

PR #1186, opened 2020-06-24
Update conflict constraint for "nelmio/api-doc-bundle" in order to allow version 3

phansys:
## Subject
Update conflict constraint for "nelmio/api-doc-bundle" in order to allow version 3.

## Changelog

```markdown
### Added
- Added `Sonata\UserBundle\Controller\Api\Legacy\GroupController` and `Sonata\UserBundle\Controller\Api\Legacy\UserController` in order to keep compatibility with nelmio/api-doc-bundle < 3.0.
- Added `SonataUserBundle/Resources/config/routing/api_forward_compatible.xml` to provide routes with nelmio/api-doc-bundle 3.x.

### Changed
- Changed "conflict" constraint for "nelmio/api-doc-bundle" in order to allow version 3.

### Deprecated
- Deprecated `Sonata\UserBundle\Controller\Api\Legacy\GroupController` and `Sonata\UserBundle\Controller\Api\Legacy\UserController`.
```

This PR should be merged after #1185 in order to leverage the functional test added there.

### DIFF
--- a/src/Controller/Api/GroupController.php
+++ b/src/Controller/Api/GroupController.php
@@ -19,9 +19,11 @@ use FOS\RestBundle\Controller\Annotations\View;
 use FOS\RestBundle\Request\ParamFetcherInterface;
 use FOS\RestBundle\View\View as FOSRestView;
 use FOS\UserBundle\Model\GroupInterface;
-use Nelmio\ApiDocBundle\Annotation\ApiDoc;
+use Nelmio\ApiDocBundle\Annotation\Model;
+use Nelmio\ApiDocBundle\Annotation\Operation;
 use Sonata\DatagridBundle\Pager\PagerInterface;
 use Sonata\UserBundle\Model\GroupManagerInterface;
+use Swagger\Annotations as SWG;
 use Symfony\Component\Form\FormFactoryInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
@@ -55,9 +57,14 @@ class GroupController
     /**
      * Returns a paginated list of groups.
      *
-     * @ApiDoc(
-     *  resource=true,
-     *  output={"class"="Sonata\DatagridBundle\Pager\PagerInterface", "groups"={"sonata_api_read"}}
+     * @Operation(
+     *     operationId="getGroups",
+     *     summary="Returns a paginated list of groups.",
+     *     @SWG\Response(
+     *         description="Returned when successful",
+     *         response="200",
+     *         @SWG\Schema(ref=@Model(type=Sonata\DatagridBundle\Pager\PagerInterface::class, groups={"sonata_api_read"}))
+     *     )
      * )
      *
      * @QueryParam(name="page", requirements="\d+", default="1", description="Page for groups list pagination (1-indexed)")
@@ -98,15 +105,25 @@ class GroupController
     /**
      * Retrieves a specific group.
      *
-     * @ApiDoc(
-     *  requirements={
-     *      {"name"="id", "dataType"="string", "description"="Group identifier"}
-     *  },
-     *  output={"class"="FOS\UserBundle\Model\GroupInterface", "groups"={"sonata_api_read"}},
-     *  statusCodes={
-     *      200="Returned when successful",
-     *      404="Returned when group is not found"
-     *  }
+     * @Operation(
+     *     operationId="getGroup",
+     *     summary="Retrieves a specific group.",
+     *     @SWG\Parameter(
+     *         name="id",
+     *         in="path",
+     *         description="Group identifier",
+     *         required=true,
+     *         type="string"
+     *     ),
+     *     @SWG\Response(
+     *         response="200",
+     *         description="Returned when successful",
+     *         @SWG\Schema(ref=@Model(type=Sonata\UserBundle\Model\GroupInterface::class, groups={"sonata_api_read"}))
+     *     ),
+     *     @SWG\Response(
+     *         response="404",
+     *         description="Returned when group is not found"
+     *     )
      * )
      *
      * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
@@ -123,13 +140,24 @@ class GroupController
     /**
      * Adds a group.
      *
-     * @ApiDoc(
-     *  input={"class"="sonata_user_api_form_group", "name"="", "groups"={"sonata_api_write"}},
-     *  output={"class"="Sonata\UserBundle\Model\Group", "groups"={"sonata_api_read"}},
-     *  statusCodes={
-     *      200="Returned when successful",
-     *      400="Returned when an error has occurred while group creation",
-     *  }
+     * @Operation(
+     *     operationId="postGroup",
+     *     summary="Adds a group.",
+     *     @SWG\Parameter(
+     *         name="",
+     *         in="body",
+     *         required=true,
+     *         @Model(type=Sonata\UserBundle\Form\Type\ApiGroupType::class, groups={"sonata_api_write"})
+     *     ),
+     *     @SWG\Response(
+     *         response="200",
+     *         description="Returned when successful",
+     *         @SWG\Schema(ref=@Model(type=Sonata\UserBundle\Model\Group::class, groups={"sonata_api_read"}))
+     *     ),
+     *     @SWG\Response(
+     *         response="400",
+     *         description="Returned when an error has occurred while group creation"
+     *     )
      * )
      *
      * @param Request $request A Symfony request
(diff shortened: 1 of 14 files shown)


phansys:
@wbloszyk, could you please check if this works fine in your environment?

wbloszyk:
I checking it in sandbox. Can't do it without other bundles upgrade.
 
IMHO we should keep support for "nelmio/api-doc-bundle" version 2 too. Mybe it is possible to add doc annotation to interface or some decorators?

phansys:
>I checking it in sandbox. Can't do it without other bundles upgrade.

Great! :+1: 

>IMHO we should keep support for "nelmio/api-doc-bundle" version 2 too. Mybe it is possible to add doc annotation to interface or some decorators?

Personally, I'd not invest time on the support for 2.x version, since it's not maintained. Do you have some strong reason to push this proposal?
Users who need support for the version 2.x can still be able to use the previous versions of this package.

wbloszyk:
> Users who need support for the version 2.x can still be able to use the previous versions of this package.

So support for 2.x can be dropped. Can you upgrade it for NewsBundle? Than this change can be checking in sandbox.
```
  Problem 1
    - nelmio/api-doc-bundle v3.6.0 conflicts with sonata-project/news-bundle[3.x-dev].
    - nelmio/api-doc-bundle v3.6.0 conflicts with sonata-project/news-bundle[3.x-dev].
    - Installation request for nelmio/api-doc-bundle 3.6 as 2.13.3 -> satisfiable by nelmio/api-doc-bundle[v3.6.0].
    - Installation request for sonata-project/news-bundle 3.x-dev -> satisfiable by sonata-project/news-bundle[3.x-dev].
```

phansys:
Here you are: sonata-project/SonataNewsBundle#604.
BTW, since I'm not user of that package, I'd prefer somebody else can be in charge to fix the TODO list in that PR.
Please, note that only the constraints were updated, in order to skip the conflict you've mentioned.

wbloszyk:
It is working. Tomorrow I will prepare sandbox for API test and issue in dev-kit with work to do.

wbloszyk:
@phansys I will be little busy in this weekend. For now you can use this: https://github.com/sonata-project/dev-kit/issues/697#issuecomment-643139293 with your NewsBundle branch.

src/Controller/Api/GroupController.php
```diff
@@ -23,9 +23,11 @@
 use FOS\RestBundle\Request\ParamFetcherInterface;
 use FOS\RestBundle\View\View as FOSRestView;
 use FOS\UserBundle\Model\GroupInterface;
-use Nelmio\ApiDocBundle\Annotation\ApiDoc;
+use Nelmio\ApiDocBundle\Annotation\Model;
+use Nelmio\ApiDocBundle\Annotation\Operation;
 use Sonata\DatagridBundle\Pager\PagerInterface;
 use Sonata\UserBundle\Model\GroupManagerInterface;
+use Swagger\Annotations as SWG;
```
core23, line 30: Why not using `Swagger` as an alias or is this best practice?

wbloszyk:
@phansys Could you change resource to FQNC [here](https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/Resources/config/routing/api.xml#L3-L4)

src/Controller/Api/GroupController.php
wbloszyk: Becouse `nelmio/api-doc-bundle` contains command `SwaggerDocblockConvertCommand` which convert ApiDoc annotations to Swagger-PHP annotations. `SWG` is usage there: 
https://github.com/nelmio/NelmioApiDocBundle/blob/master/UPGRADE-3.0.md

wbloszyk:
@phansys We should keep support for `nelmio/api-doc-bundle` v2 as long as v3 is not added to whole sonata. I add PR how we can do it in easy way:
https://github.com/phansys/SonataUserBundle/pull/1

phansys:
Thank you @wbloszyk.
I've merged the PR and then I made some changes. Please, let me know if they seem fine for you.

jordisala1991:
I think having another folder is not a great idea, and we should consider that nelmio 4 changes another time all the annotations... another folder?

phansys:
>I think having another folder is not a great idea, and we should consider that nelmio 4 changes another time all the annotations... another folder?

I'd prefer avoid supporting "nelmio/api-doc-bundle:4.x" in parallel with a previous version, since it's intended to work with [OAS 3.x](https://swagger.io/specification/), while the versions 3.x and 2.x are for [OAS 2.x](https://swagger.io/specification/v2/).
Additionally, I think supporting both versions in this PR is not mandatory either. We are making the best effort to avoid forcing all the `sonata-project/*` packages to make this change in sync (https://github.com/sonata-project/SonataUserBundle/pull/1186#issuecomment-657415790).

wbloszyk:
IMHO, mybe we should add Base*ApiController and make sturcure like that?:
```
- Controller\Api
    - BaseUserController
    - Nelmio2UserController
    - Nelmio3UserController
```
Support for nelmio 2 can be keep in sonata 3 then (with deprecation notes). People can add support for Nelmio 4 byself in easy or mybe evan we can do it? WDYT? @phansys @jordisala1991

phansys:
I really don't like the idea to create an extension point in controllers just for the OAS definition.
I know currently, the only way to provide the required metadata with NelmioApiDocBundle is through annotations (the case could be other if a decoupled mechanism were supported, like XML or YAML), but I think in our case that will demand too much effort with no huge core benefit.
Even, I'd say the users should be in charge to choose which OAS version they want to implement. Obviously, we still can provide a default implementation, ideally with the latest available version, and document how to provide the specification in a different version for these controllers (inheriting, overriding, etc).

src/Controller/Api/GroupController.php
```diff
@@ -62,9 +64,14 @@ public function __construct(GroupManagerInterface $groupManager, FormFactoryInte
     /**
      * Returns a paginated list of groups.
      *
-     * @ApiDoc(
-     *  resource=true,
-     *  output={"class"="Sonata\DatagridBundle\Pager\PagerInterface", "groups"={"sonata_api_read"}}
+     * @Operation(
```
wbloszyk, line 67:
```suggestion
     * @SWG\Operation(
```
Can we use `@SWG` instead add another FQNC?

src/Controller/Api/Legacy/GroupController.php
```diff
@@ -0,0 +1,275 @@
...
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Controller\Api\Legacy;
+
+@trigger_error(
+    'The '.__NAMESPACE__.'\GroupController class is deprecated since sonata-project/user-bundle 4.x and will be removed in version 5.0.'
```
wbloszyk, line 17:
```suggestion
    'The '.__NAMESPACE__.'\GroupController class is deprecated since sonata-project/user-bundle 4.x and will be removed in 5.0.'
```
It is nesessery? Mybe we should add templates for deprecations to keep some standards? @phansys WDYT? You are the best in it.

src/Controller/Api/Legacy/GroupController.php
```diff
@@ -0,0 +1,275 @@
...
+     * @ApiDoc(
+     *  input={"class"="sonata_user_api_form_group", "name"="", "groups"={"sonata_api_write"}},
+     *  output={"class"="Sonata\UserBundle\Model\Group", "groups"={"sonata_api_read"}},
+     *  statusCodes={
+     *      200="Returned when successful",
+     *      400="Returned when an error has occurred while group creation",
+     *  }
+     * )
+     *
+     * @Post("/group")
```
wbloszyk, line 151:
```suggestion
     * @REST\Post("/group")
```
The same here for `@REST`.

src/Controller/Api/GroupController.php
phansys: I guess we could, but I'm not sure we should. Regardless the `SWG` alias was already imported, IMO we should avoid to propagate the usage of classes which are provided by packages which are not declared explicitly as direct dependencies.

nelmio/NelmioApiDocBundle#973.

src/Controller/Api/Legacy/GroupController.php
phansys: >It is nesessery?

I'm afraid yes, since is the only way I know to tell users that this class will be not available in the next major release.

>Mybe we should add templates for deprecations to keep some standards? @phansys WDYT?

I think we could use [symfony/deprecation-contracts](https://github.com/symfony/deprecation-contracts) as starting point.

>You are the best in it.

I'm sorry to disagree, but surely I'm not the best. Regarding deprecations, exceptions, etc, I'm only putting the focus in leveraging the message structures and the wording already established by the language and other packages in the ecosystem. Trying to be as semantic, explicit and clear as possible.

src/Controller/Api/Legacy/GroupController.php
phansys: Done.

wbloszyk:
WDYT about more complex tests? @phansys  
https://github.com/sonata-project/SonataClassificationBundle/pull/545/files#diff-2f1fac9c29233575ae3896af487090ee

phansys:
> WDYT about more complex tests? @phansys

The would be the goal, but the last time I tried there were some dependencies or missing model records preventing the API endpoints from giving successful status codes. That's why I've opted for narrow the tests only to the routing purposes, regardless what its controllers return.

wbloszyk:
@phansys Could you rebase it?

phansys:
> @phansys Could you rebase it?

Done.

wbloszyk:
Closed in favor of #1225

PR #1195, opened 2020-07-24
Fix passing "format" parameter to API routes

phansys:
## Subject
Fix passing "format" parameter to API routes.

Related to https://github.com/sonata-project/SonataMediaBundle/pull/1767#issuecomment-663496518.

## Changelog

```markdown
### Fixed
- Fixed passing "format' parameter to API routes.
```

### DIFF
--- a/src/Controller/Api/UserController.php
+++ b/src/Controller/Api/UserController.php
@@ -32,9 +32,12 @@ use Symfony\Component\Form\FormFactoryInterface;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Routing\Annotation\Route;
 
 /**
  * @author Hugo Briand <briand@ekino.com>
+ *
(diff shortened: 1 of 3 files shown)


src/Controller/Api/UserController.php
```diff
@@ -68,7 +71,7 @@ public function __construct(UserManagerInterface $userManager, GroupManagerInter
      *  output={"class"="Sonata\DatagridBundle\Pager\PagerInterface", "groups"={"sonata_api_read"}}
      * )
      *
-     * @Get("/users", name="get_users")
+     * @Get("/users.{_format}", name="get_users")
```
wbloszyk, line 74:
```suggestion
     * @Get("/users.{_format}", name="get_users", requirements={"_format": "json|xml|html"})
```
I check it and compare with [demo](https://demo.sonata-project.org/). I think we should add requirements for it too. I don't know what when people set rules for it or move some format. WDYT? @phansys

phansys: I've added the requirements globally in each controller via `@Route` annotation in order to share the definition for all the routes in that scope.

jordisala1991:
Thank you @phansys

PR #1196, opened 2020-07-24
Use service ids instead of controller paths at API routing config definitions

phansys:
## Subject
Use service ids instead of controller paths at API routing config definitions.

## Changelog

```markdown
### Changed
- Changed API routing config definitions in order to use service ids instead of controller paths.
```


VincentLanglet:
Thanks @phansys

PR #1198, opened 2020-07-27
Update Swagger definition in order to be compatible with string identifiers

phansys:
## Subject
Update OpenAPI (Swagger) definition in order to be compatible with string identifiers (like UUIDs).

These changes are consistent with the API narrowing made at sonata-project/admin-bundle (see [`AdminInterface::id()`](https://github.com/sonata-project/SonataAdminBundle/blob/09a61607afb13eb36db56c11bdc05a497c17d684/src/Admin/AdminInterface.php#L206)).

## Changelog

```markdown
### Removed
- Removed requirements that were only allowing integers for model identifiers at Open API definitions.
### Fixed
- Fixed support for string model identifiers at Open API definitions.
```

### DIFF
--- a/src/Controller/Api/GroupController.php
+++ b/src/Controller/Api/GroupController.php
@@ -100,7 +100,7 @@ class GroupController
      *
      * @ApiDoc(
      *  requirements={
-     *      {"name"="id", "dataType"="integer", "requirement"="\d+", "description"="group id"}
+     *      {"name"="id", "dataType"="string", "description"="Group identifier"}
      *  },
      *  output={"class"="FOS\UserBundle\Model\GroupInterface", "groups"={"sonata_api_read"}},
      *  statusCodes={
@@ -111,7 +111,7 @@ class GroupController
      *
      * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
-     * @param $id
+     * @param string $id
      *
      * @return GroupInterface
      */
@@ -148,7 +148,7 @@ class GroupController
      *
      * @ApiDoc(
      *  requirements={
-     *      {"name"="id", "dataType"="integer", "requirement"="\d+", "description"="group identifier"}
+     *      {"name"="id", "dataType"="string", "description"="Group identifier"}
      *  },
      *  input={"class"="sonata_user_api_form_group", "name"="", "groups"={"sonata_api_write"}},
      *  output={"class"="Sonata\UserBundle\Model\Group", "groups"={"sonata_api_read"}},
@@ -159,7 +159,7 @@ class GroupController
      *  }
      * )
      *
-     * @param int     $id      Group identifier
+     * @param string  $id      Group identifier
      * @param Request $request A Symfony request
      *
      * @throws NotFoundHttpException
@@ -176,7 +176,7 @@ class GroupController
      *
      * @ApiDoc(
      *  requirements={
-     *      {"name"="id", "dataType"="integer", "requirement"="\d+", "description"="group identifier"}
+     *      {"name"="id", "dataType"="string", "description"="Group identifier"}
      *  },
      *  statusCodes={
      *      200="Returned when group is successfully deleted",
@@ -185,7 +185,7 @@ class GroupController
      *  }
      * )
      *
-     * @param int $id A Group identifier
+     * @param string $id A Group identifier
      *
      * @throws NotFoundHttpException
      *
@@ -203,8 +203,8 @@ class GroupController
     /**
      * Write a Group, this method is used by both POST and PUT action methods.
      *
-     * @param Request  $request Symfony request
-     * @param int|null $id      A Group identifier
+     * @param Request     $request Symfony request
+     * @param string|null $id      A Group identifier
      *
      * @return FormInterface
      */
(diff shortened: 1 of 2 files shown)


phansys:
@wbloszyk, could you please confirm if the sandbox is working fine without these restrictions?

wbloszyk:
> @wbloszyk, could you please confirm if the sandbox is working fine without these restrictions?

In `@ApiDoc` this restriction in only like information for people. It will be work but you should keep `"dataType"="integer|UUIDs"`. Restriction will be when you add it in `@Route` and `@ApiDoc` will read this and display descriptions for people from it.

phansys:
>In @ApiDoc this restriction in only like information for people. 

AFAIK, the @ApiDoc annotations are used to build the Open API Specification, which is consumed by the built-in sandbox to display the information which you are referencing and to create the "Try it out" forms. It also can be consumed by 3rd party compatible clients.

phansys:
If you can, please check the resulting definition at [Swagger Editor](https://editor.swagger.io/). You can also play with the changes you're proposing in order to check if the result works as expected.
>It will be work but you should keep "dataType"="integer|UUIDs"

I wouldn't use "UUID" as type, since their type is really a string.

src/Controller/Api/UserController.php
```diff
@@ -178,8 +178,8 @@ public function postUserAction(Request $request)
      *
      * @Put("/user/(id}.{_format}", name="put_user")
```
wbloszyk, line 179:
```suggestion
     * @Put("/user/{id}.{_format}", name="put_user")
```
Can you check in other places too?

wbloszyk:
For swagger for path parameter you have to:
```php
@SWG\Parameter(name="id", in="path", type="string")
```
Type must be "string", "number", "integer", "boolean", "array", "file". 
"string" works becouse it is convert to url but probably it is not good idea. 
"integer" will not allow use string.

src/Controller/Api/UserController.php
phansys: Good catch! Thank you!
sonata-project/SonataUserBundle#1202.

wbloszyk: This will be remove in https://github.com/sonata-project/SonataUserBundle/pull/1200

phansys: >This will be remove in #1200

Yes.

phansys:
>"string" works becouse it is convert to url but probably it is not good idea.

Why do you think is not a good idea? IMO, we could use `type="string"` in conjunction with [`pattern`](https://github.com/GuilhemN/swagger/blob/a02984db5edacdce2b4e09dae5ba8fe17a0e449e/src/Parts/TypePart.php#L315) attribute in order to keep narrowed the allowed values.

src/Controller/Api/GroupController.php
```diff
@@ -100,7 +100,7 @@ public function getGroupsAction(ParamFetcherInterface $paramFetcher)
      *
      * @ApiDoc(
      *  requirements={
-     *      {"name"="id", "dataType"="integer", "requirement"="\d+", "description"="group id"}
+     *      {"name"="id", "dataType"="string", "description"="group id"}
```
wbloszyk, line 103:
```suggestion
     *      {"name"="id", "dataType"="string", "description"="Group identifier"}
```
Can we improve description too?

src/Controller/Api/GroupController.php
```diff
@@ -111,7 +111,7 @@ public function getGroupsAction(ParamFetcherInterface $paramFetcher)
      *
      * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
-     * @param $id
+     * @param string $id
```
wbloszyk, line 114:
```suggestion
     * @param string|int $id Group identifier
```
I think we should we should add `int`, otherwise someone can add `string` type hint. Also Can we repeat desciption here?

src/Controller/Api/GroupController.php
```diff
@@ -159,7 +159,7 @@ public function postGroupAction(Request $request)
      *  }
      * )
      *
-     * @param int     $id      Group identifier
+     * @param string  $id      Group identifier
      * @param Request $request A Symfony request
```
wbloszyk, line 163:
```suggestion
     * @param Request $request Symfony request
```
I think we can avoid using `the` or `a` in field description. It's look better in documentation.

src/Controller/Api/GroupController.php
phansys: >I think we should we should add int, otherwise someone can add string type hint.

Parameter types can be certainly almost anything you want given the way the coercion works from the type declarations.

```
curl http://app/test/42/42/42/42
```
```php
/**
 * @Route("/test/{idUntyped}/{idString}/{idInteger}/{idFloat}")
 */
public function getTestAction($untyped, string $idString, int $idInteger, float $idFloat)
{
    var_dump($idUntyped); // string(2) "42"
    var_dump($idString); // string(2) "42"
    var_dump($idInteger); // int(42)
    var_dump($idFloat); // float(42)
}
```

Given that, I think we SHOULD declare these parameters as `string` in the next major.

>Also Can we repeat desciption here?

I'd prefer to let this task for a separate PR.

src/Controller/Api/GroupController.php
phansys: I've updated the suggested descriptions at Swagger declarations. Let's these docblock description changes for a separate PR.

wbloszyk (approved):

wbloszyk:
Thanks @phansys

PR #1200, opened 2020-07-27
Use XML for ReST API routing

phansys:
## Subject
Use XML for ReST API routing.

Related to https://github.com/sonata-project/SonataMediaBundle/pull/1767#discussion_r460857425.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,6 @@
         "matthiasnoback/symfony-config-test": "^4.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
         "nelmio/api-doc-bundle": "^2.4",
-        "sensio/framework-extra-bundle": "^5.5",
         "sonata-project/doctrine-orm-admin-bundle": "^3.18",
         "sonata-project/google-authenticator": "^1.0 || ^2.0",
         "symfony/browser-kit": "^4.4 || ^5.1",
--- a/src/Controller/Api/GroupController.php
+++ b/src/Controller/Api/GroupController.php
@@ -14,10 +14,6 @@ declare(strict_types=1);
 namespace Sonata\UserBundle\Controller\Api;
 
 use FOS\RestBundle\Context\Context;
-use FOS\RestBundle\Controller\Annotations\Delete;
-use FOS\RestBundle\Controller\Annotations\Get;
-use FOS\RestBundle\Controller\Annotations\Post;
-use FOS\RestBundle\Controller\Annotations\Put;
 use FOS\RestBundle\Controller\Annotations\QueryParam;
 use FOS\RestBundle\Controller\Annotations\View;
 use FOS\RestBundle\Request\ParamFetcherInterface;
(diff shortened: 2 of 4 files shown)


wbloszyk:
https://github.com/sonata-project/SonataMediaBundle/pull/1767#discussion_r460957196 @phansys

jordisala1991:
Even if it is not required, isn't it better to just use the xml format anyway? With this, if someday the route changes we don't have to duplicate the controller to not break BC (like in the nelmio case).

Bonus: all the other routes are defined like this. wdyt @wbloszyk ? IMO it is worth it from every point of view.

jordisala1991:
I don't use the api at all, I didn't check if everything works. Are the api endpoints + docs generated correctly now?

jordisala1991:
Another point: You have to pick one: It is pedantic and no changelog or it is a patch/minor then it deserves a changelog.

phansys:
>I don't use the api at all, I didn't check if everything works. Are the api endpoints + docs generated correctly now?

I'm not using the API either, but since the tests are checking the routes are generated as expected, I think there is no problem.
https://github.com/sonata-project/SonataUserBundle/blob/cf1f385bb06ed35194400410dceb6ad5b621a3da/tests/Functional/Routing/RoutingTest.php#L22
Swagger doc should not be affected, since it's agnostic about how routing is defined.

>Another point: You have to pick one: It is pedantic and no changelog or it is a patch/minor then it deserves a changelog.

IMO, this change is pedantic. I'm just filling the changelog for internal reference.

jordisala1991:
The thing with the changelog is that is auto generated now , so if we dont remove it, it will end up in the release notes

wbloszyk:
> Even if it is not required, isn't it better to just use the xml format anyway? With this, if someday the route changes we don't have to duplicate the controller to not break BC (like in the nelmio case).
> 
> Bonus: all the other routes are defined like this. wdyt @wbloszyk ? IMO it is worth it from every point of view.

IMHO annotations is good in small close/ecommerce projects. In Sonata using xml for Rest Api, like in other places, will be awsome.    

> I don't use the api at all, I didn't check if everything works. Are the api endpoints + docs generated correctly now?

I use it from time to time, but not so much complex. I compare it with [demo](https://demo.sonata-project.org/) to keep it the same. We (with @phansys) added smoke tests for it and have add functional tests in plan.

> Another point: You have to pick one: It is pedantic and no changelog or it is a patch/minor then it deserves a changelog.

IMHO it is BC-break for people who override API. I think it is minor and little upgrade note should be add.

wbloszyk:
I update https://github.com/sonata-project/dev-kit/issues/778. We should know how far we are in work on Rest API and what we do.

jordisala1991:
Given the point that since not long ago the api wasnt working at all, I would not consider this as a Bc break

phansys:
>IMHO it is BC-break for people who override API. I think it is minor and little upgrade note should be add. Otherwise they will have two diffrence system (self annotations and sonata xml) for routing.

AFAIK, the only way users have to override the API routing is using their own routing definitions referencing to the existing controllers. If that is the case, I think there is no BC break at all.

>Given the point that since not long ago the api wasnt working at all, I would not consider this as a Bc break

I agree. Even, the routing annotations were added just in the last release ([4.6.0](https://github.com/sonata-project/SonataUserBundle/releases/tag/4.6.0)).

phansys:
Changelog removed in order to be picked up by release process. It was:
## Changelog






```markdown
### Changed
- Changed routing definition from annotations to XML.
### Removed
- Removed dev requirement for "sensio/framework-extra-bundle".
```

jordisala1991:
Thank you @phansys

PR #1201, opened 2020-07-27
Fix deprecation messages in `AdminResettingController` and `AdminResettingController`

phansys:
## Subject
Fix deprecation messages in `AdminResettingController` and `AdminResettingController`, since currently they are making references to non existing classes.

### DIFF
--- a/tests/Controller/AdminResettingControllerTest.php
+++ b/tests/Controller/AdminResettingControllerTest.php
@@ -75,7 +75,7 @@ class AdminResettingControllerTest extends TestCase
 
     /**
      * @group legacy
-     * @expectedDeprecation The Sonata\UserBundle\Controller\AdminResettingController class is deprecated since version 4.3.0 and will be removed in 5.0. Use Sonata\UserBundle\Controller\RequestAction, Sonata\UserBundle\Controller\CheckEmailAction, Sonata\UserBundle\Controller\ResetAction or Sonata\UserBundle\Controller\SendEmailAction instead.
+     * @expectedDeprecation The Sonata\UserBundle\Controller\AdminResettingController class is deprecated since version 4.3.0 and will be removed in 5.0. Use Sonata\UserBundle\Action\RequestAction, Sonata\UserBundle\Action\CheckEmailAction, Sonata\UserBundle\Action\ResetAction or Sonata\UserBundle\Action\SendEmailAction instead.
      */
     public function testCheckEmailAction(): void
     {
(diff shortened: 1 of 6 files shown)


jordisala1991:
There might me something wrong on tests:

```
Since symfony/phpunit-bridge 5.1: Using "@expectedDeprecation" annotations in tests is deprecated, use the "ExpectDeprecationTrait::expectDeprecation()" method instead.
```

seems like we using the old way of expecting deprecations. Deprecated 😅

phansys:
>There might me something wrong on tests:

I've replaced those annotations. The test should be fine now.

tests/Controller/AdminResettingControllerTest.php
```diff
@@ -20,12 +20,15 @@
...
 use Sonata\UserBundle\Action\SendEmailAction;
 use Sonata\UserBundle\Controller\AdminResettingController;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 
 class AdminResettingControllerTest extends TestCase
 {
+    use ExpectDeprecationTrait;
```
jordisala1991, line 30: I think this is only needed if you using really old versions of phpunit. on 8.5 we should be fine without it

phansys: I can confirm it is required. We are using phpunit/phpunit:8.5.0:
https://github.com/sonata-project/SonataUserBundle/blob/d60215ea263d81ee2e40ff9dc2df795c6ca1d299/phpunit.xml.dist#L35

Please, see symfony/symfony#37153.

jordisala1991: Oh I see, needed then, I was confused with the expectException

VincentLanglet:
There should be no changelog, for pedantic isn't it ?

And I think two commit would be better.
- One with the deprecation message fix
- One removing the deprecated `@expectedDeprecation`. (Not sure if it was really needed btw, since we'll remove all of them in the next major ; all the other SonataBundle use this annotation, so now we're in a not-consistent situation 😅)

phansys:
Changelog section removed, since these changes are pedantic. It was:
## Changelog






```markdown
### Fixed
- Fixed QCNs referenced by deprecation messages at `AdminResettingController` and `AdminResettingController`;
- Fixed deprecated usage of `@exptectedDeprecation` through `ExpectDeprecationTrait::expectDeprecation()`.
```

phansys:
>There should be no changelog, for pedantic isn't it ?

Changelog removed.

>And I think two commit would be better.

Commits splitted.

jordisala1991:
Thank you @phansys

PR #1205, opened 2020-08-08
Remove oppressive words from code

### LINKED ISSUE #1191: Remove oppressive words from code
## Feature Request

We should avoid terms like *blacklist*, *whitliste, *master*, *slave*, ... and use more tech-related words like *blocklist*, *prmary*, *secondary*,...

In this repo, we use the term *whitelist* which could be replaced with *allowlist*:
https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/GoogleAuthenticator/Helper.php#L42

davidromani:
## Subject

We should avoid offending terms like blacklist, whitlist, master, slave... and use more tech-related words like blocklist, primary or secondary, for example.

I am targeting 4.x branch because everything is backwards compatible.

Closes #1191 

## Changelog

```markdown
### Changed
- Replace `ip_white_list` by `trusted_ip_list` configuration key
```

## To do

- [x] Update the tests;
- [x] Update the documentation;
- [x] Add an upgrade note.


docs/reference/two_step_validation.rst
```diff
@@ -35,12 +35,12 @@ Edit the configuration file:
         google_authenticator:
             enabled: true
             server:  yourserver.com
-            ip_white_list:
+            trusted_ip_list:
```
greg0ire, line 38: You could make this BC by adding a new configuration key, and deprecate the old one in favor of the new one, which will make upgrading easier.

davidromani:
Ok, let me a little bit more time, I prefer to make it BC too...

davidromani:
Hey @greg0ire I don't know how to change this PR against 4.x branch instead of master branch. Is this possible? Or I must open a new PR?

franmomu:
> Hey @greg0ire I don't know how to change this PR against 4.x branch instead of master branch. Is this possible? Or I must open a new PR?

You can click on the Edit button on the top right (like changing the title) and then change the base branch.

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -178,7 +178,7 @@ public function configureGoogleAuthenticator($config, ContainerBuilder $containe
         }
 
         $container->setParameter('sonata.user.google.authenticator.forced_for_role', $config['google_authenticator']['forced_for_role']);
-        $container->setParameter('sonata.user.google.authenticator.ip_white_list', $config['google_authenticator']['ip_white_list']);
+        $container->setParameter('sonata.user.google.authenticator.trusted_ip_list', $config['google_authenticator']['trusted_ip_list']);
```
franmomu, line 181: This would be BC-break, people can still use the `ip_white_list` option, so here we should also check if `ip_white_list` exists and use it. A test verifying that this behavior still works would be nice.

davidromani: @franmomu I've marked `ip_white_list` as a deprecated parameter, is this enough?

src/DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,9 @@ public function getConfigTreeBuilder()
                         ->scalarNode('server')->cannotBeEmpty()->end()
                         ->scalarNode('enabled')->defaultFalse()->end()
                         ->arrayNode('ip_white_list')
+                            ->setDeprecated('The "%node%" option is deprecated. Use "trusted_ip_list" instead with the same values.')
```
davidromani, line 72: @greg0ire @franmomu is this enough?

franmomu: For the deprecation it is, sorry, I meant if it was possible to deprecate a parameter in the container since someone can be using the `sonata.user.google.authenticator.ip_white_list` parameter.

src/DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,9 @@ public function getConfigTreeBuilder()
...
+                        ->end()
+                        ->arrayNode('trusted_ip_list')
```
greg0ire, line 74: It would be great to throw an exception if both configuration nodes are defined. I'm not sure if it can be done here, if yes, that's great, if not you can do it in the extension.

src/DependencyInjection/SonataUserExtension.php
greg0ire: `ip_white_list` is a configuration node. A container parameter is something else entirely. AFAIK You cannot deprecate it, but maybe I'm wrong. IMO both parameters from whatever configuration node is defined in the config (after validating that one and only one of them is defined as I described above).

franmomu: What I meant is removing `sonata.user.google.authenticator.ip_white_list` from the container is BC-break because someone can be using that parameter (it's unlikely, but you never know), if there is no way to deprecate a parameter in the container, an upgrade note I think it's enough and we will remove it on the next major version.

franmomu:
So just to be clear because I didn't explain myself clearly, there are three issues right now:
- A user can define `ip_white_list` and `trusted_ip_list`, in that case as @greg0ire said it would be great to throw an exception.
- In case someone defines `ip_white_list`, they will get a deprecation message, but it won't work right now, we should do something like:
```php
$trustedIpList = $config['google_authenticator']['ip_white_list'] ?? $config['google_authenticator']['trusted_ip_list'];
$container->setParameter('sonata.user.google.authenticator.trusted_ip_list', trustedIpList);
```
- We shouldn't remove `sonata.user.google.authenticator.ip_white_list` parameter from the container because is BC-break, so we should add:
```php
$container->setParameter('sonata.user.google.authenticator.ip_white_list', trustedIpList);
```

davidromani:
@franmomu @greg0ire only as a proof of concept, is the last commit in a good path to solve it?

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -178,7 +178,15 @@ public function configureGoogleAuthenticator($config, ContainerBuilder $containe
...
 
         $container->setParameter('sonata.user.google.authenticator.forced_for_role', $config['google_authenticator']['forced_for_role']);
-        $container->setParameter('sonata.user.google.authenticator.ip_white_list', $config['google_authenticator']['ip_white_list']);
+
+        if (array_key_exists('ip_white_list', $config['google_authenticator']) && array_key_exists('trusted_ip_list', $config['google_authenticator'])) {
+            throw new \RuntimeException('Please use only ``trusted_ip_list`` parameter. ``ip_white_list`` is deprecated.');
+        }
+        if (array_key_exists('ip_white_list', $config['google_authenticator']) && !array_key_exists('trusted_ip_list', $config['google_authenticator'])) {
+            $container->setParameter('sonata.user.google.authenticator.trusted_ip_list', $config['google_authenticator']['ip_white_list']);
+        } else {
```
VincentLanglet, line 187: You should add a comment
`// NEXT_MAJOR: Remove this checks and only set the trusted_ip_list.`

That's how we deal todos.

src/DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,12 @@ public function getConfigTreeBuilder()
                         ->scalarNode('server')->cannotBeEmpty()->end()
                         ->scalarNode('enabled')->defaultFalse()->end()
                         ->arrayNode('ip_white_list')
+                            ->prototype('scalar')->end()
+                            ->defaultValue(['127.0.0.1'])
+                            ->info('IPs for which 2FA will be skipped. DEPRECATED: replace by "trusted_ip_list" instead parameter with the same values')
+                            ->setDeprecated()
```
greg0ire, line 75: I'm confused… why is `setDeprecated` empty now?

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -178,7 +178,15 @@ public function configureGoogleAuthenticator($config, ContainerBuilder $containe
         }
 
         $container->setParameter('sonata.user.google.authenticator.forced_for_role', $config['google_authenticator']['forced_for_role']);
-        $container->setParameter('sonata.user.google.authenticator.ip_white_list', $config['google_authenticator']['ip_white_list']);
+
+        if (array_key_exists('ip_white_list', $config['google_authenticator']) && array_key_exists('trusted_ip_list', $config['google_authenticator'])) {
+            throw new \RuntimeException('Please use only ``trusted_ip_list`` parameter. ``ip_white_list`` is deprecated.');
```
greg0ire, line 183: This should be fixed by changing code, so let's use `\LogicException` here
```suggestion
            throw new \LogicException('Please use only ``trusted_ip_list`` parameter. ``ip_white_list`` is deprecated.');
```

greg0ire: Also, why the double backticks?

greg0ire:
> @franmomu @greg0ire only as a proof of concept, is the last commit in a good path to solve it?

It's looking good yeah :)

src/DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,12 @@ public function getConfigTreeBuilder()
                         ->scalarNode('server')->cannotBeEmpty()->end()
                         ->scalarNode('enabled')->defaultFalse()->end()
                         ->arrayNode('ip_white_list')
+                            ->prototype('scalar')->end()
+                            ->defaultValue(['127.0.0.1'])
```
franmomu, line 73: If there is a default value in both options, I guess you would always get an exception.

VincentLanglet: If we provide a default value for one option, we will still have an exception when the other will be defined.

I'm not sure the exception is the right thing to do.
We could
- Remove the ipwhitelist default value
- If ipwhitelist is defined, use it. Else use trustediplist.

src/DependencyInjection/SonataUserExtension.php
davidromani: Done

src/DependencyInjection/Configuration.php
davidromani: Fixed

src/DependencyInjection/SonataUserExtension.php
davidromani: Seen in line 177 of the same class, I've tried to keep same notation (don't know if must be single or double backtick)

src/DependencyInjection/Configuration.php
davidromani: Solved

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -178,7 +178,17 @@ public function configureGoogleAuthenticator($config, ContainerBuilder $containe
         }
 
         $container->setParameter('sonata.user.google.authenticator.forced_for_role', $config['google_authenticator']['forced_for_role']);
-        $container->setParameter('sonata.user.google.authenticator.ip_white_list', $config['google_authenticator']['ip_white_list']);
+
+        // NEXT_MAJOR: Remove this checks and only set the `trusted_ip_list`.
+        if (\array_key_exists('ip_white_list', $config['google_authenticator']) && \array_key_exists('trusted_ip_list', $config['google_authenticator'])) {
```
franmomu, line 183: `$config['google_authenticator']['ip_white_list']` and `$config['google_authenticator']['trusted_ip_list']` are always defined, there is no need to check with `array_key_exists`.

Given that, you could do something like:
```php
if (count($config['google_authenticator']['ip_white_list']) > 0 && $config['google_authenticator']['trusted_ip_list'] !== ['127.0.0.1']) {
    throw new \LogicException('Please use only ``trusted_ip_list`` parameter. ``ip_white_list`` is deprecated.');
}
```

The idea is that `trusted_ip_list` will always have `['127.0.0.1']` value since it's the default value specified in `Configuration`. 

And then you could do something like:

```php
$trustedIpList = $config['google_authenticator']['trusted_ip_list'];

if (\count($config['google_authenticator']['ip_white_list']) > 0) {
    $trustedIpList = $config['google_authenticator']['ip_white_list'];
}

$container->setParameter('sonata.user.google.authenticator.ip_white_list', $trustedIpList);
$container->setParameter('sonata.user.google.authenticator.trusted_ip_list', $trustedIpList);
```

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -42,7 +42,7 @@ public function testDefault(): void
             ],
             'google_authenticator' => [
                 'enabled' => false,
-                'ip_white_list' => ['127.0.0.1'],
+                'trusted_ip_list' => ['127.0.0.1'],
```
franmomu, line 45:
```suggestion
                'ip_white_list' => [],
                'trusted_ip_list' => ['127.0.0.1'],
```

Since `ip_white_list` is still accepted, it will be in the default values, with this the test should be fixed.

davidromani: Fixed

src/DependencyInjection/SonataUserExtension.php
greg0ire: This is neither markdown nor rst, don't know why it was contributed like this. IMO we should use double quotes…

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -178,7 +178,18 @@ public function configureGoogleAuthenticator($config, ContainerBuilder $containe
         }
 
         $container->setParameter('sonata.user.google.authenticator.forced_for_role', $config['google_authenticator']['forced_for_role']);
-        $container->setParameter('sonata.user.google.authenticator.ip_white_list', $config['google_authenticator']['ip_white_list']);
+
+        // NEXT_MAJOR: Remove this checks and only set the `trusted_ip_list`.
+        if (\count($config['google_authenticator']['ip_white_list']) > 0 && $config['google_authenticator']['trusted_ip_list'] !== ['127.0.0.1']) {
+            throw new \LogicException('Please use only ``trusted_ip_list`` parameter, ``ip_white_list`` is deprecated.');
```
greg0ire, line 184:
```suggestion
            throw new \LogicException('Please use only "trusted_ip_list" parameter, "ip_white_list" is deprecated.');
```

greg0ire (approved):

jordisala1991 (approved):
Can you please squash to a single commit ? If you dont know how to do it, we need to make sure we do a squash merge

davidromani:
> Can you please squash to a single commit ? If you dont know how to do it, we need to make sure we do a squash merge

Never done before, but I think that I can do it... let me try it

greg0ire:
Here is a guide:

1. `git rebase -i origin/4.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

davidromani:
> Looks like you did it! One last thing: please format your commit message according to our rules :

All done, thanks for your patience and for teaching me how to improve my Git skills.

franmomu (approved):

franmomu:
Thanks @davidromani!

PR #1212, opened 2020-08-26
Removed config line causing issue on google authenticator

Jean-ita:
## Subject
This fixes the following error:
Stop sonata-project/google-authenticator throwing this exception: Exception thrown when handling an exception (ErrorException: Warning: get_class() expects parameter 1 to be object, null given at /sonata-project/user-bundle/src/GoogleAuthenticator/RequestListener.php line 77

## Changelog

```markdown

### Fixed
- Removed sonata.templating argument from google_authenticator.xml
```

### DIFF
--- a/src/Resources/config/google_authenticator.xml
+++ b/src/Resources/config/google_authenticator.xml
@@ -19,8 +19,6 @@
             <tag name="kernel.event_listener" event="kernel.request" method="onCoreRequest" priority="-1"/>
             <argument type="service" id="sonata.user.google.authenticator.provider"/>
             <argument type="service" id="security.token_storage"/>
-            <!-- NEXT_MAJOR: Remove "sonata.templating" argument -->
-            <argument type="service" id="sonata.templating" on-invalid="null"/>
             <argument type="service" id="twig"/>
         </service>
         <service id="sonata.user.google.authenticator.success_handler" class="Sonata\UserBundle\EventListener\TwoFactorLoginSuccessHandler" public="false">


Jean-ita:
@jordisala1991 this is what we talked about on the sonata slack channel

jordisala1991:
There are 2 ways to fix the issue, the first one is to remove that line, IMO the cleanest one.

The other way is to allow null on the construct (now it is forbidden with an Exception here: https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/GoogleAuthenticator/RequestListener.php#L72

I think what happened is that we removed the `sonata.templating` (or it was on the CoreBundle?). So you could end up in the `on-invalid=null` case.

jordisala1991:
Seems like your github account and your git config is not the same, that causes your commit to not show your avatar. Can you link your email to your git config?

Jean-ita:
> Seems like your github account and your git config is not the same, that causes your commit to not show your avatar. Can you link your email to your git config?

Done. Next commit will show the correct avatar

phansys (approved):
IMO, the changelog should use the "Fixed" section instead of "Changed", since this is a patch.

Jean-ita:
> IMO, the changelog should use the "Fixed" section instead of "Changed", since this is a patch.

True, it's done 👍

VincentLanglet (approved):

VincentLanglet:
Thanks @Jean-ita

Issue #1112, opened 2019-08-20
Remove FOSUserBundle inheritance


core23:
## Feature Request

This bundle is dead: https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2874#issuecomment-462644756

There are some threads that were asking for new versions:
- https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2927 
- https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2917 
- https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2874

So we should remove the coupling to this bundle and fork/implement some of the basic functions here.

grimabe:
Regarding the lack of maintenance of FosUserBundle and Symfony 5 release approaching this issue is getting more urgent. I don't know if there is a roadmap at Sonata somewhere but I am curious to see if there is any news about this issue ?
Thanks

core23:
This is not a topic we're working on at the moment. Our top prio is to release new major versions of our bundles. 

Maybe you can help on this topic @grimabe ?

core23:
FYI there is a new bundle which is a hard fork of the FOSUserBundle with some major cleanup: 
https://github.com/nucleos/NucleosUserBundle

I will create some additional bundles which will add a frontend profile and registration

VincentLanglet:
Do we need these new bundle ? 

I understood in the FOSUser issue that they recommend to use the maker-bundle and this documentation https://symfony.com/doc/current/security.html

core23:
> Do we need these new bundle ?
> 
> I understood in the FOSUser issue that they recommend to use the maker-bundle and this documentation https://symfony.com/doc/current/security.html

Not sure if we need user bundles anymore, but I don't like the "there is only one way for a user provider" approach that symfony does. It will work for most cases, but there is no alternative way anymore.

haivala:
I have been digging this for a while now and seems that there is many ways to do this but I just need to ask this: What is the recommended way to upgrade a project with SonataUserBundle not to use FOSUserBundle so that it is future proof? I'm using Symfony 4.4.

VincentLanglet:
> upgrade a project with SonataUserBundle not to use FOSUserBundle so that it is future proof

I'm not sure you can as long as SonataUserBundle use FOSUserBundle.
PR should be done first on this project to remove the dependency.

haivala:
So what you are saying is that the maker-bundle way does not work with SonataUserBundle?

[...]

PR #1250, opened 2020-10-10
FosUserBundle require doctrine/common 2

VincentLanglet:
Tests are currently failing because it install doctrine common 3 and doctrine persistence 2.

But FOSUserBundle is requiring doctrine common 2.

Until FosUser is removed, cf https://github.com/sonata-project/SonataUserBundle/issues/1112, this fix the tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
     ],
     "require": {
         "php": "^7.2",
+        "doctrine/common": "^2.0",
         "friendsofsymfony/user-bundle": "^2.0",
         "sonata-project/admin-bundle": "^3.34",
         "sonata-project/datagrid-bundle": "^3.0.1",


jordisala1991 (approved):

jordisala1991:
Thank you @VincentLanglet , I was thinking if it is not better to use the conflict section instead

VincentLanglet:
> Thank you @VincentLanglet , I was thinking if it is not better to use the conflict section instead

I should be required by FosUser. But it's not in their composer.json.
Since they abandonned the project, we have to add it in our dependencies IMHO.

jordisala1991:
> > Thank you @VincentLanglet , I was thinking if it is not better to use the conflict section instead
> 
> 
> 
> I should be required by FosUser. But it's not in their composer.json.
> 
> Since they abandonned the project, we have to add it in our dependencies IMHO.

Agree, This is another thing we have to resolve for Sonata 4 (well this will be SonataUserBundle 5)

PR #1254, opened 2020-10-13
Stop extending `AbstractAdmin::getExportFields()`

phansys:
## Subject
Stop extending `AbstractAdmin::getExportFields()`, since this method is marked as final.

## Changelog

```markdown
### Fixed
- Deprecation about the extension of `AbstractAdmin::getExportFields()` method in `UserAdmin`.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "php": "^7.2",
         "doctrine/common": "^2.0",
         "friendsofsymfony/user-bundle": "^2.0",
-        "sonata-project/admin-bundle": "^3.34",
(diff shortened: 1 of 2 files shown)


greg0ire (approved):
It looks like this also will have the side effect of making extensions added to the `UserAdmin` able to change export fields. Link for the lazy: https://github.com/sonata-project/SonataAdminBundle/blob/6d32a6fc4e285dfe6ae32af6d2428d122e33e9c9/src/Admin/AbstractAdmin.php#L665-L676

jordisala1991:
Thank you @phansys

PR #1260, opened 2020-10-21
Fix: PHPUnit 9 compatibility

OskarStark:
`The at() matcher has been deprecated. It will be removed in PHPUnit 10. Please refactor your test to not rely on the order in which methods are invoked.`


jordisala1991 (approved):

jordisala1991:
Thank you @OskarStark

PR #1264, opened 2020-10-26
4.9.0

VincentLanglet:


wbloszyk (changes requested):
We should not remove UPGRADE-4.0.md

VincentLanglet:
> We should not remove UPGRADE-4.0.md

The upgrade note is 3 year old.
And if someone want the upgrade note, he just have to look at https://github.com/sonata-project/SonataUserBundle/tree/4.0.0.

wbloszyk (approved):

VincentLanglet:
> IMO It should be deleted in 4.1.0(4.0.1)

I agree, better late than never

PR #1256, opened 2020-10-15
Drop FOSUserBundle with minimum functionality

wbloszyk:
## Subject

This PR move minimul functionality from FOSUserBundle which will be allow using SonataUserBundle without FOSUserBundle dependencies/inheritance.

I am targeting this branch, because this change are BC-BREAK.

Closes #1255

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

## To do

- [x] Update the tests;
- [x] Update the documentation;
- [ ] Add an upgrade note.

PR #1281, opened 2020-11-17
Don't break 2FA when mixing user-providers

erik-zicht:
## Don't break 2FA when mixing user-providers

`getTwoStepVerificationCode` is not always present on every user-object, e.g. the in-memory user (https://github.com/symfony/security-core/blob/5.x/User/User.php) does not have this method. 
The current implementation makes it impossible to mix different user-providers.

I am targeting this branch, because it's a patch.

## Changelog

```markdown
### Changed
- `TwoFactorLoginSuccessHandler` verifies if 2FA business-logic can be executed.
```

### DIFF
--- a/src/EventListener/TwoFactorLoginSuccessHandler.php
+++ b/src/EventListener/TwoFactorLoginSuccessHandler.php
@@ -72,7 +72,7 @@ final class TwoFactorLoginSuccessHandler implements AuthenticationSuccessHandler
         $user = $token->getUser();
         $needToHave2FA = $this->googleAuthenticator->needToHaveGoogle2FACode($request);
 
-        if ($needToHave2FA && !$user->getTwoStepVerificationCode()) {
+        if ($needToHave2FA && method_exists($user, 'getTwoStepVerificationCode') && !$user->getTwoStepVerificationCode()) {
             $secret = $this->googleAuthenticator->generateSecret();
             $user->setTwoStepVerificationCode($secret);
 
@@ -88,7 +88,7 @@ final class TwoFactorLoginSuccessHandler implements AuthenticationSuccessHandler
                     'error' => [],
                 ]
             ));
-        } elseif ($needToHave2FA && $user->getTwoStepVerificationCode()) {
+        } elseif ($needToHave2FA && method_exists($user, 'getTwoStepVerificationCode') && $user->getTwoStepVerificationCode()) {
             $request->getSession()->set($this->googleAuthenticator->getSessionKey($token), null);
         }
 


VincentLanglet (changes requested):
I see multiple issues here:
- This bundle is currently made for FosUser ; so it's not suppose to work with custom User/UserProvider. Changing this would be a lot of work: https://github.com/sonata-project/SonataUserBundle/pull/1256

- For what I understand, there is a specific interface to use https://github.com/sonata-project/SonataUserBundle/blob/16ff16cfb17d16b7117c9afd6cde59ac7f0e6384/src/Model/UserInterface.php with the `getTwoStepVerificationCode` method.

- An `instanceof` check should be preferred to `method_exists($user, 'getTwoStepVerificationCode')`

- The check is not made anywhere https://github.com/sonata-project/SonataUserBundle/search?q=getTwoStepVerificationCode.

- In this case if `method_exists($user, 'getTwoStepVerificationCode')` does not exists, both `if` and `elseif` will be false so why not setting `needToHave2FA` to `false` instead ?

erik-zicht:
Ah, perfect, I wasn't aware a specific interface existed. 

With your other comments, would this PR be sufficient when I introduce an `instanceof` check and rewrite the conditions a bit?

VincentLanglet:
> With your other comments, would this PR be sufficient when I introduce an `instanceof` check and rewrite the conditions a bit?

I'm not sure. I think more context could be great.

So, how do you use this bundle ?

The doc is describing the following config
```
# config/packages/sonata_user.yaml

sonata_user:
    class:
        user: App\Entity\SonataUserUser
        group: App\Entity\SonataUserGroup
```
I assume you used your own User class instead ?

The `instanceof UserInterface` check seems to restrictive.

IMHO the right solution would be:
- Creating an interface `FooInterface` (with an appropriate name) with the method `getTwoStepVerificationCode` and `setTwoStepVerificationCode`.
- Removing the method from the `UserInterface` and extending FooInterface, `UserInterface extends FooInterface`.
- Making the check on `FooInterface`.

This way it will be easier for developers to implement the interface.

But then, If I look at the code, it's
```
$needToHave2FA = $this->googleAuthenticator->needToHaveGoogle2FACode($request);

if ($needToHave2FA && $user instanceof FooInterface && ...) {
 ...
} elseif ($needToHave2FA && $user instanceof FooInterface && ...) {
 ...
}
```
If `$needToHave2FA` is `true`, which I assume is the case for you (since you have the error). 
I don't understand how the code could be valid if you don't pass by the `if` or the `elseif`.

So I think it should be instead
```
if ($needToHave2FA) {
    if (! $user instanceof FooInterface) {
        throw new \Exception('User must implement FooInterface`);
    }

    if (...) {
       ....
    } elseif ( ...) {
        ...
    }
}
```

And then you'll need to implement FooInterface.

Because how the 2FA could work without a `TwoStepVerificationCode` ?

erik-zicht:
```
security: 
   providers:
        chain_provider:
            chain:
                providers: [ in_memory, fos_userbundle ]
```

My use-case is a project in which I have a mix of in-memory users and database-users.
When logging in with a in-memory user, there can never be need for 2FA, as there would not be a place to store any information. For all other users there must be, as it is a functional requirement that we have to meet. If both conditions are negative I'm only redirected to `$url`, basically bypassing all the code that is in this class.

I'm not looking to do a rewrite of the existing code outside of the `TwoFactorLoginSuccessHandler`. My patch is focused on bypassing the 2FA code when not needed, without causing errors.

I agree with all your pointers, introducing extra interfaces and such, (I've had experience before with this one - works exactly how you describe https://github.com/scheb/2fa/blob/5.x/src/google-authenticator/Model/Google/TwoFactorInterface.php) but I'm not fixing that right now. 

Feel free to close the PR and report a new issue on how to rewrite this in the best possible way, if desired.

VincentLanglet:
> When logging in with a in-memory user, there can never be need for 2FA, as there would not be a place to store any information. 

Then there is some logic issues: `$needToHave2FA` shouldn't be true, if you don't want 2FA.

> I'm not looking to do a rewrite of the existing code outside of the `TwoFactorLoginSuccessHandler`. My patch is focused on bypassing the 2FA code when not needed, without causing errors.
> I agree with all your pointers, introducing extra interfaces and such, (I've had experience before with this one - works exactly how you describe https://github.com/scheb/2fa/blob/5.x/src/google-authenticator/Model/Google/TwoFactorInterface.php) but I'm not fixing that right now.

You patch is focused on your special use case, getting errors because you may not use this handler the right way.
That's I'm trying to find, on my spare time.

This library is used by lot of persons, with different use case.
What if someone comes next asking why there is no 2FA, with absolutely no error, when it's needed (`$needToHave2FA = true`) ?

> Feel free to close the PR and report a new issue on how to rewrite this in the best possible way, if desired.

Seems like this issue is not so important for you...

erik-zicht:
I'll close the PR, I think there is no immediate patch possible and this should be addressed in a broader manner.

PR #1426, opened 2021-10-30
Remove nelmio v4 from conflict

wbloszyk:
## Subject

I am targeting this branch, because this change should be done here.

## Changelog

```markdown
### Removed
- Removed conflict with `nelmio/api-doc-bundle` v4
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
     "suggest": {
         "friendsofsymfony/rest-bundle": "For using the public API methods.",
         "jms/serializer": "For using the public API methods.",
-        "nelmio/api-doc-bundle": "v2 or v3 - for using the public API methods. v4 does not provide API.",
+        "nelmio/api-doc-bundle": "Versions 2.x or 3.x for using the public ReST API methods. Version 4.x is allowed but not supported.",
         "sonata-project/doctrine-orm-admin-bundle": "For persisting entities",
         "sonata-project/google-authenticator": "For google auth user login"
     },
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
     "conflict": {
         "friendsofsymfony/rest-bundle": "<2.1",
         "jms/serializer": "<1.3 || >=4.0",
-        "nelmio/api-doc-bundle": "<2.13 || >=4.0",
+        "nelmio/api-doc-bundle": "<2.13 || >=5.0",
         "sonata-project/core-bundle": "<3.20",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -80,7 +80,7 @@
     "suggest": {
         "friendsofsymfony/rest-bundle": "For using the public API methods.",
         "jms/serializer": "For using the public API methods.",
-        "nelmio/api-doc-bundle": "For using the public API methods.",
+        "nelmio/api-doc-bundle": "v2 or v3 - for using the public API methods. v4 does not provide API.",
```
phansys:
```suggestion
        "nelmio/api-doc-bundle": "Versions 2.x or 3.x for using the public ReST API methods. Version 4.x is allowed but not supported.",
```

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -88,7 +89,9 @@ class_exists('Sonata\GoogleAuthenticator\GoogleAuthenticator')) {
             $loader->load('serializer.xml');
 
             $loader->load('api_form.xml');
-            if (class_exists(Operation::class)) {
+            if (class_exists(Operation::class) && class_exists(OpenApiOperation::class)) {
+                // UserBundle does not provide API for nelmio v4. This code allow remove nelmio v4 from conflict.
```
phansys:
```suggestion
                // UserBundle does not support "nelmio/api-doc-bundle" 4.x and does not provide an API through this version. This condition allows to accept setups using "nelmio/api-doc-bundle" 4.x dependency without conflicts.
                // @no-op.
```

phansys (changes requested):

wbloszyk:
In cases when people using API and havent defind nelmio version in own composer then v4 will be installed. Aby idea haw resole it?
@phansys @sonata-project/contributors

phansys:
>In cases when people using API and havent defind nelmio version in own composer then v4 will be installed. Aby idea haw resole it?

Why the package will be installed in that case? IIUC, in your example the package is not required by the userland project, and neither is it required by "sonata-project/user-bundle".
Or do you mean the userland project requiring something like `"sonata-project/user-bundle": "*"`?
If it is the case, I think is responsibility of the user to get rid about the resolved dependencies when performing an update, as the problem there is the loose constraint.

VincentLanglet (approved):

phansys (approved):

VincentLanglet:
Thanks

PR #1430, opened 2021-11-04
Update login.html.twig

AntoineRoue:
## Subject

Taking into account the following change : "title_mode" (with values "single_image", "single_text", "both") is deprecated since sonata-project/admin-bundle 3.104 and replaced by "logo_content" (with values "icon", "text", "all").

I'm not sure if I should've added a comment "{# NEXT_MAJOR: Remove the title_mode check #}" because it is for the next major of SonataAdminBundle.

## Changelog
```markdown
### Fixed
- Taking into account sonata_admin.options.logo_content configuration value
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "doctrine/common": "^2.0 || ^3.1",
         "doctrine/persistence": "^1.3 || ^2.1",
         "friendsofsymfony/user-bundle": "^2.0",
-        "sonata-project/admin-bundle": "^3.76",
+        "sonata-project/admin-bundle": "^3.104",
         "sonata-project/datagrid-bundle": "^3.0.1",
         "sonata-project/doctrine-extensions": "^1.10.1",
         "sonata-project/form-extensions": "^0.1 || ^1.4",
--- a/src/Resources/views/Admin/Security/Resetting/checkEmail.html.twig
+++ b/src/Resources/views/Admin/Security/Resetting/checkEmail.html.twig
@@ -28,13 +28,19 @@ file that was distributed with this source code.
         {% block login_box_header %}
             <div class="login-logo">
                 <a href="{{ path('sonata_admin_dashboard') }}">
-                    {% if admin_pool.getOption('title_mode') in ['single_image', 'both'] %}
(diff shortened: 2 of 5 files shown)


jordisala1991:
You should also update minimum version of the admin bundle, because previous versions do not have this code deprecated

AntoineRoue:
You mean setting `"sonata-project/admin-bundle": "^3.104",` in composer.json, instead of ^3.76 currently ? If so, I'm not sure we can do this easily.

jordisala1991:
Yes, And we are able to do this, its not a bc break to update a dependency, you just have to change the version on the composer.json

jordisala1991:
I started to work on 5.x branch to prepare for SF 5 / 6, without FOSUser and all. Please take a look at how I solve this on tihs precise PR to copy the same code: https://github.com/sonata-project/SonataUserBundle/pull/1444

This will help with the merge to higher branches.

AntoineRoue:
Do you want me to update checkEmail, request and reset files as well in this PR ?

Compared to my changes, you removed the conditions ```and ('single_text' == sonata_config.getOption('title_mode') or 'both' == sonata_config.getOption('title_mode'))```, isn't it ? However, as the option title_mode is deprecated but not yet removed, how can you not take title_mode value into account ?
If title_mode configuration is changed to single_text or single_image, with your change it will always display both logo and text.

jordisala1991:
You're right, my PR is only for Sonata 4, where I do not take into account the old deprecated values, because they are removed on Sonata 4. And yes, it would be nice to fix those 3 templates at once, it should share the same fix.

AntoineRoue:
Finished my copy paste.

VincentLanglet (approved):

jordisala1991 (approved):

VincentLanglet:
@AntoineRoue Can you add a changelog in your first message to explain the purpose of this PR. This will be used for the next release. You can take a look at other PR to see the format.

AntoineRoue:
Done. Is that good ?

VincentLanglet:
> Done. Is that good ?

Thanks

PR #1437, opened 2021-11-28
DevKit updates for 4.x branch

SonataCI:


jordisala1991:
I would say it is not worth try to add php 8.1 here with the current situation.

VincentLanglet:
> I would say it is not worth try to add php 8.1 here with the current situation.

According to the errors, I'm not sure it would be hard.
But sure, we can delay it.

First thing would be to solve the Sf5/Sonata4 issue ; which require a lot of work...

jordisala1991:
> > I would say it is not worth try to add php 8.1 here with the current situation.
> 
> 
> 
> According to the errors, I'm not sure it would be hard.
> 
> But sure, we can delay it.
> 
> 
> 
> First thing would be to solve the Sf5/Sonata4 issue ; which require a lot of work...

I am not sure how to solve the errors, do you want to give them a try? Otherwise I revert the php 8.1

VincentLanglet:
@jordisala1991 The 8.1 build is solved.
Only psalm is failing. I think it would require to add `returnTypeWillChange` on SonataAdmin 3.
Dunno if we want to do this.

jordisala1991:
Not to me, IMO we should focus on 5.x of this bundle. But if you want to invest time, lets do it.

admin bundle and block bundle 3.x should probably stop being on dev-kit and stop being maintained.

What do you say, we revert or we continue?

VincentLanglet:
We can revert

VincentLanglet (approved):

PR #1438, opened 2021-12-01
Bump PHP

wbloszyk:
## Subject

I am targeting this branch, because this change respect BC.

## Changelog

```markdown
### Removed
- Support for php 7.3
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,6 @@ jobs:
         strategy:
             matrix:
                 php-version:
-                    - '7.3'
                     - '7.4'
                     - '8.0'
                 dependencies: [highest]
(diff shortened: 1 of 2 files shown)


VincentLanglet:
See https://github.com/sonata-project/SonataUserBundle/pull/1437, we may need a PR to SonataAdmin3 for the 8.1 support.

wbloszyk:
> See #1437, we may need a PR to SonataAdmin3 for the 8.1 support.

So lets remove support for 7.3 only.

VincentLanglet:
This require a PR on devkit (to remove ci on 7.3 and revert the fact we're requiring 8.1 for the ci)

jordisala1991:
IMO it does not make  a lot of sense to do it on 4.x, there is a little benefit since you won't be able to add typehints and this branch won't be able to upgrade much more (because of FOS). IMO the efforts should be on 5.x  branch. I already started with it. It needs a lot of cleanup to prepare for the FOSUser removal. (I will do it but I need some time).

PR #1455, opened 2021-12-17
[POC] Add symfonycasts/reset-password-bundle

wbloszyk:
## Subject

POC for replace FOSUser resetting password into symfonycasts resetting password.

I am targeting this branch, because {reason}.

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
         "symfony/security-acl": "^3.0",
         "symfony/security-core": "^4.4.23",
         "symfony/translation": "^4.4",
+        "symfonycasts/reset-password-bundle": "^1.11",
         "twig/twig": "^2.9 || ^3.0"
     },
     "require-dev": {
--- a/src/Action/CheckEmailAction.php
+++ b/src/Action/CheckEmailAction.php
@@ -21,6 +21,12 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Twig\Environment;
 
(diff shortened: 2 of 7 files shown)


jordisala1991:
This has several problems that I solved already on RunroomUserBundle. We can go into details if you want, but I recommend you to go to that bundle, check the code and see the problem we faced on that integration (It is not suported to reset password for more of 1 entity, if you want to make it optional you have to deal with the entities without using the interface directly ).

But again , why is that Pr related to the FOSUser removal? It can and should be handled separately because of the mentioned issues and because it is NOT needed to remove FOSUser.

Hanmac:
> This has several problems that I solved already on RunroomUserBundle. We can go into details if you want, but I recommend you to go to that bundle, check the code and see the problem we faced on that integration (It is not suported to reset password for more of 1 entity, if you want to make it optional you have to deal with the entities without using the interface directly ).

yeah i noticed that too, (my newest project needs password reset for multiple user entities)
so i made an Issue there: https://github.com/SymfonyCasts/reset-password-bundle/issues/203

i mean you can make it possible if you have the Reset Provider branch for different User and ResetToken classes and delegate that stuff.

wbloszyk:
I use it in multiple user project but i use BaseUser and discriminator for it.

VincentLanglet:
Closing since we're gonna release the RC

PR #1457, opened 2021-12-17
Simplify Actions code

jordisala1991:
## Subject

Not sure if changelog is needed, the behavior didn't change but some deps where not used so...

### DIFF
--- a/src/Action/LoginAction.php
+++ b/src/Action/LoginAction.php
@@ -21,10 +21,8 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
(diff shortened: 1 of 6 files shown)


Hanmac:
@jordisala1991 just a reminder, you changed the `$csrfTokenManager` from a setter to a constructor parameter, you need to update the new php config too

was there maybe a reason why it was a setter in the FOSUser one?

i was thinking if it could depend on the setting in the form_login security thing, but i don't think that can be accessed in the controller

jordisala1991:
> @jordisala1991 just a reminder, you changed the `$csrfTokenManager` from a setter to a constructor parameter, you need to update the new php config too
> 
> was there maybe a reason why it was a setter in the FOSUser one?
> 
> i was thinking if it could depend on the setting in the form_login security thing, but i don't think that can be accessed in the controller

I removed the setter because it can be done with the constructor too, to me it looks cleaner. We might try to use the setting of the form_login configuration, but that will be for another PR.

VincentLanglet (approved):

PR #1468, opened 2022-01-03
Remove groups feature

jordisala1991:
## Subject

To me it does not need extra deprecations on 4.x since FOSUser already deprecated the feature and we are relying on it.

## Changelog

```markdown
### Removed
- Removed user groups feature.
```


src/Admin/Model/UserAdmin.php
```diff
@@ -87,9 +81,8 @@ protected function configureFormFields(FormMapper $form): void
             ->end()
             ->tab('Security')
                 ->with('Status', ['class' => 'col-md-4'])->end()
-                ->with('Groups', ['class' => 'col-md-4'])->end()
                 ->with('Keys', ['class' => 'col-md-4'])->end()
-                ->with('Roles', ['class' => 'col-md-12'])->end()
+                ->with('Roles', ['class' => 'col-md-16'])->end()
```
VincentLanglet: This is wrong. 12 is the max. You should instead change the col-md-4 to col-md-6 for status and keys.

VincentLanglet (changes requested):

src/Admin/Model/UserAdmin.php
jordisala1991: Tried to fix it on my last commit. Should be better now.

VincentLanglet (approved):

PR #1325, opened 2021-02-08
Add support for "doctrine/common:^3"

phansys:
## Subject
Add support for "doctrine/common:^3".

I am targeting this branch, because these changes respect BC.

Closes #1317

## Changelog

```markdown
### Added
- Support for "doctrine/common:^3".
```

## To do:
- [x] Check if more class aliases must be provided.

PR #1475, opened 2022-01-04
DevKit updates for 4.x branch

SonataCI:


VincentLanglet:
Certainly related to https://github.com/sonata-project/SonataUserBundle/pull/1325 @jordisala1991

jordisala1991:
> Certainly related to https://github.com/sonata-project/SonataUserBundle/pull/1325 @jordisala1991 

Yes, we added some kind of hack that the lint container does not detect. But the fosUser should be compatible now, maybe removing the hack helps, or we might need to disable this checks here.

VincentLanglet:
I think it's ok to disable it for now

PR #1476, opened 2022-01-08
Fix serialize method

VincentLanglet:
## Subject

As reported by @Jean-ita, the serialize method should was wrong.

## Changelog

```markdown
### Fixed
- `User::__serialize()` and `User::__unserialize()` methods
```

### DIFF
--- a/src/Model/User.php
+++ b/src/Model/User.php
@@ -145,7 +145,7 @@ abstract class User extends AbstractedUser implements UserInterface
      */
     public function __serialize()
     {
-        return serialize([
+        return [
             $this->password,
             $this->salt,
             $this->usernameCanonical,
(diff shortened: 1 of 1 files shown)


jordisala1991 (dismissed):
Can you also check unserialize? I think I got it right on 5.x so it should be a good example.

VincentLanglet:
> Can you also check unserialize? I think I got it right on 5.x so it should be a good example.

I updated the __unserialize method too.

Basically `serialize( SonataUser::__serialize) === FosUser::serialize` and same for unserialize (if you understand the syntax) ^^'.

Since we're still depending on FosUser, which have some BC code in the unserialize method, I prefered to keep the BC code too ; dunno what's for.

jordisala1991 (approved):

PR #1479, opened 2022-01-09
Resolve NEXT_MAJOR comments

jordisala1991:
## Subject

I am targeting this branch, because this is for next major.

## Changelog

```markdown
### Added
- Added `EditableRolesBuilderInterface`.

### Changed
- Changed `__construct()` of `EditableRolesBuilder` to provide `TranslatorInterface`
```

### DIFF
--- a/src/Security/EditableRolesBuilder.php
+++ b/src/Security/EditableRolesBuilder.php
@@ -19,7 +19,7 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
-final class EditableRolesBuilder
+final class EditableRolesBuilder implements EditableRolesBuilderInterface
 {
     private TokenStorageInterface $tokenStorage;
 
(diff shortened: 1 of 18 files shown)


VincentLanglet:
Tests need to be updated
- Some private methods were tested
- Some mock were made on final classes

VincentLanglet:
For the label, I think the `major` label require to have a Changelog.
So as a hack, we have to use the pedantic label in order to generate the release changelog with the dev kit command.

jordisala1991:
> For the label, I think the `major` label require to have a Changelog. So as a hack, we have to use the pedantic label in order to generate the release changelog with the dev kit command.

I have added the changelog, had to introduce a new interface.

I didn't touch a metod that should be private because I will be fixing that on another PR. (I noted it on the 5.0 roadmap issue)

src/Security/EditableRolesBuilder.php
```diff
@@ -42,23 +39,22 @@ class EditableRolesBuilder
...
      * @param array<string, array<string>> $rolesHierarchy
      */
-    public function __construct(TokenStorageInterface $tokenStorage, AuthorizationCheckerInterface $authorizationChecker, Pool $pool, SonataConfiguration $configuration, array $rolesHierarchy = [])
-    {
+    public function __construct(
+        TokenStorageInterface $tokenStorage,
+        AuthorizationCheckerInterface $authorizationChecker,
+        Pool $pool,
+        SonataConfiguration $configuration,
+        ?TranslatorInterface $translator = null,
```
VincentLanglet: Translator can be removed isnt it ? It seems not used

VincentLanglet (changes requested):

src/Security/EditableRolesBuilder.php
jordisala1991: Is it used, check below.

VincentLanglet: Indeed, on the last lines.

Test need to be updated and this should be ok.

jordisala1991: Now, they should be okay.

VincentLanglet (dismissed):

VincentLanglet (approved):

Issue #1465, opened 2021-12-27
Roadmap to 5.0


jordisala1991:
## Feature Request

There are more things to do for 5.0:

- [x] Avoid using parameters on container as much as possible
- [x] User form types instead of hardcoded forms
- [x] Html support in email
- [x] Fix UserAdmin only change password (add tests similar to MediaBundle)
- [x] Add commands (adapt commands from FOSUserBundle)
- [x] Psalm level 2
- [x] Deprecate/Remove Groups feature
- [x] Update documentation
- [x] Improve configuration #1464 https://github.com/sonata-project/SonataUserBundle/pull/1481
- [x] Mark final classes
- [x] Move from public/protected to private
- [x] Avoid direct usage of container https://github.com/sonata-project/SonataUserBundle/pull/1481
- [x] Simplify impersonating routing configuration https://github.com/sonata-project/SonataUserBundle/pull/1481
- [x] Check twig templates
- [x] Replace glyphicons by font awesome icons
- [x] Improve tests (mark final, remove legacy)
- [x] Use array instead of json for roles

Translations:
- [x] Add missing translations for english
- [x] Add missing translations for rest of languages (FOSUser)
- [x] Remove untranslated translations
- [x] Add missing validation translations (FOSUser)

Tests:
- [x] Test migration from SonataUser v4 to v5 (previous user data works) (4.4)
- [x] Test all features from the bundle on new security system (6.0)
- [x] Test all features from the bundle on the old security system (5.4)

Documentation:
- [x] Improve documentation for Symfony 5/6 installs

As soon as I know more things to do I will keep adding them.

andyhobbs:
Will be Groups feature in another bundle?

jordisala1991:
If you just need to group some roles you can use roles_hierarchy from symfony already.

If you need some more complex solution, you might take a look at voters. The groups feature was not a good solution since it forces always a query to check for the groups roles, it wont be provided by Sonta (you can still reimplement it by yourself if you really need to)

jordisala1991:
After #1490 we are almost there

Things missing:
* Fixing translations (we have to migrate some translations from FOSUser on other languages, and check if there are placeholder translations and things like that)
* Improving documentation for Symfony 5 / 6 (security system is different so we might provide configuration with both Symfony versions)

[...]

PR #1484, opened 2022-01-27
Add commands from FOSUserBundle

jordisala1991:
## Subject

I am targeting this branch, because here is where commands are not present because of removal of FOSUserBundle.

Part of #1465 

## Changelog

```markdown
### Added
- Added `sonata:user:create` command from FOSUserBundle
- Added `sonata:user:promote` command from FOSUserBundle
- Added `sonata:user:demote` command from FOSUserBundle
- Added `sonata:user:change-password` command from FOSUserBundle
- Added `sonata:user:activate` command from FOSUserBundle
- Added `sonata:user:deactivate` command from FOSUserBundle
```


src/Util/UserManipulator.php
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Sonata\UserBundle\Util;
+
+use Sonata\UserBundle\Model\UserInterface;
+use Sonata\UserBundle\Model\UserManagerInterface;
+
+/**
+ * @internal
+ */
+final class UserManipulator
```
jordisala1991: Not sure if this class makes sense or it is better to use the UserManager we already have, thoughts @VincentLanglet ?

VincentLanglet: Do we need all these methods ? You're only using `activate` in your command

src/Command/ActivateUserCommand.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        parent::__construct();
+
+        $this->userManipulator = $userManipulator;
+    }
+
+    protected function configure()
+    {
+        $this
+            ->setName('fos:user:activate')
```
core23:
```suggestion
            ->setName('sonata:user:activate')
```

src/Util/UserManipulator.php
jordisala1991: There are more commands coming, we will use all the methods.

VincentLanglet: With the first command I would just write directly
```
$user = $this->userManager->findUserByUsername();

$user->setEnabled(true);

$this->userManager->save($user);
```
directly in the command.

src/Command/ActivateUserCommand.php
```diff
@@ -0,0 +1,83 @@
...
+            ->setDescription('Activate a user')
+            ->setDefinition([
+                new InputArgument('username', InputArgument::REQUIRED, 'The username'),
+            ])
+            ->setHelp(
+                <<<'EOT'
+The <info>fos:user:activate</info> command activates a user (so they will be able to log in):
```
dmetzner: sonata:user:activate

src/Util/UserManipulator.php
jordisala1991: Done.

src/Command/ActivateUserCommand.php
jordisala1991: Fixed.

src/Command/ActivateUserCommand.php
jordisala1991: Fixed.

src/Command/ActivateUserCommand.php
```diff
@@ -0,0 +1,95 @@
...
+        $user->setEnabled(true);
+
+        $this->userManager->save($user);
+
+        $output->writeln(sprintf('User "%s" has been activated.', $username));
+
+        return 0;
+    }
+
+    protected function interact(InputInterface $input, OutputInterface $output): void
```
jordisala1991: Do you see this useful? To me it looks not, and it didn't work in FOSUserBundle, because it requires the InputArgument to be OPTIONAL, otherwise the only way to trigger it is:

```
bin/console sonata:user:activate '' # (Does not work without an empty argument, because it is required)
```

jordisala1991: I made the InputArgument OPTIONAL in the meantime, but if we decide to remove this I will make it REQUIRED again.

VincentLanglet: I don't find it really useful. We could for the simplest command with the required argument IMHO.

jordisala1991: Removed.

tests/Functional/Command/ChangePasswordCommandTest.php
```diff
@@ -0,0 +1,114 @@
...
+        static::assertSame($user->getPassword(), 'old_password');
+
+        $this->commandTester->execute([
+            'username' => 'sonata-user-test',
+            'password' => 'new_password',
+        ]);
+
+        $user = $this->refreshUser($user);
+
+        // static::assertSame($user->getPassword(), 'new_password');
```
jordisala1991: The password is not really changed.

That's because the plainPassword field is not managed by Doctrine, so if we only modify this it is unable to know something changed and trigger the preUpdate lifecycle event.

On FOSUserBundle , the UserManager "save" method does not only save the object but also updated the canonical fields and the password. Should we do the same?

It will also affect the UserAdmin when you only change password, so we should do the same there.

wdyt @VincentLanglet ? At the end we have some duplicated things. The preUpdate event and the code on the manager and the admin.

jordisala1991: Fixed by adding the updatePassword to the ModelManager, not sure if I like it a lot but it works.

psalm-baseline.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.19.0@a2ad69ae4f5ab1f7d225a8dc4e2ec2d9415ed599">
+    <file src="src/Action/LoginAction.php">
+        <UndefinedInterfaceMethod occurrences="1">
+            <code>getFlashBag</code>
```
VincentLanglet, line 5: Soon, it won't be needed https://github.com/psalm/psalm-plugin-symfony/pull/242 ;)

VincentLanglet (approved):

jordisala1991:
Lets add this now, to not block other Prs

PR #1486, opened 2022-01-30
Mark tests final

jordisala1991:
## Subject

I am targeting this branch, because this is in preparation for 5.0.

Part of #1465


tests/Action/SendEmailActionTest.php
```diff
@@ -25,38 +25,29 @@
...
      * @var MockObject&TokenGeneratorInterface
      */
-    protected MockObject $tokenGenerator;
+    private MockObject $tokenGenerator;
 
-    protected int $resetTtl;
-
-    protected string $fromEmail;
-
-    protected string $template;
```
VincentLanglet, line 54: This was removed, and still used

VincentLanglet (changes requested):

tests/Action/SendEmailActionTest.php
jordisala1991: It was only used in the setUp function, fixed now.

VincentLanglet (approved):

PR #1487, opened 2022-01-30
Add more admin tests

jordisala1991:
## Subject

I am targeting this branch, because this is for next major.

Part of #1465

### DIFF
--- a/src/Admin/Model/UserAdmin.php
+++ b/src/Admin/Model/UserAdmin.php
@@ -80,8 +80,7 @@ class UserAdmin extends AbstractAdmin
                 ->with('General', ['class' => 'col-md-12'])->end()
             ->end()
             ->tab('Security')
-                ->with('Status', ['class' => 'col-md-6'])->end()
-                ->with('Keys', ['class' => 'col-md-6'])->end()
+                ->with('Status', ['class' => 'col-md-12'])->end()
                 ->with('Roles', ['class' => 'col-md-12'])->end()
             ->end();
 
--- a/src/Admin/Model/UserAdmin.php
+++ b/src/Admin/Model/UserAdmin.php
@@ -32,6 +32,8 @@ class UserAdmin extends AbstractAdmin
 
     public function __construct(UserManagerInterface $userManager)
     {
+        parent::__construct();
+        
         $this->userManager = $userManager;
     }
 
--- a/src/Admin/Model/UserAdmin.php
+++ b/src/Admin/Model/UserAdmin.php
@@ -20,6 +20,7 @@ use Sonata\AdminBundle\FieldDescription\FieldDescriptionInterface;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\AdminBundle\Show\ShowMapper;
 use Sonata\UserBundle\Form\Type\SecurityRolesType;
+use Sonata\UserBundle\Model\UserManagerInterface;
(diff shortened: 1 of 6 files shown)


src/Admin/Model/UserAdmin.php
```diff
@@ -20,13 +20,30 @@
...
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 
 /**
  * @phpstan-extends AbstractAdmin<\Sonata\UserBundle\Model\UserInterface>
  */
 class UserAdmin extends AbstractAdmin
 {
+    private ?UserManagerInterface $userManager;
+
+    public function setUserManager(UserManagerInterface $userManager): void
```
VincentLanglet: Wouldn't it better to use constructor ?

src/Admin/Model/UserAdmin.php
```diff
@@ -20,13 +20,30 @@
...
 use Sonata\UserBundle\Form\Type\SecurityRolesType;
+use Sonata\UserBundle\Model\UserManagerInterface;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 
 /**
  * @phpstan-extends AbstractAdmin<\Sonata\UserBundle\Model\UserInterface>
  */
 class UserAdmin extends AbstractAdmin
 {
+    private ?UserManagerInterface $userManager;
```
VincentLanglet: Shouldn't we provide a getter or make the property protected if the class is not final ?

src/Admin/Model/UserAdmin.php
jordisala1991: I set the min version to 4.8 of SonataAdmin and use the constructor.

src/Admin/Model/UserAdmin.php
jordisala1991: I made it protected.

src/Admin/Model/UserAdmin.php
```diff
@@ -20,13 +20,26 @@
...
 /**
  * @phpstan-extends AbstractAdmin<\Sonata\UserBundle\Model\UserInterface>
  */
 class UserAdmin extends AbstractAdmin
 {
+    protected UserManagerInterface $userManager;
+
+    public function __construct(UserManagerInterface $userManager)
+    {
+        $this->userManager = $userManager;
```
VincentLanglet, line 36:
```suggestion
        parent::__construct();
        
        $this->userManager = $userManager;
```

to please phpstan

jordisala1991: Added, but not sure if it is better to add to the baseline.

src/Admin/Model/UserAdmin.php
```diff
@@ -20,13 +20,28 @@
...
  * @phpstan-extends AbstractAdmin<\Sonata\UserBundle\Model\UserInterface>
  */
 class UserAdmin extends AbstractAdmin
 {
+    protected UserManagerInterface $userManager;
+
+    public function __construct(UserManagerInterface $userManager)
+    {
+        parent::__construct();
+
```
jordisala1991:
```suggestion

```

src/Admin/Model/UserAdmin.php
VincentLanglet: I think it's ok to have this "useless" call so far. It will be removed for the next sonata admin major.

VincentLanglet (approved):

PR #1489, opened 2022-01-31
Add form type for reset password request

### LINKED ISSUE #784: Use twig blocks for form rendering
We should replace all `<form>` elements in the twig templates with the twig function `form_start` and `form_end`. We should also move the `action` and `method` (POST, GET) attribute to the PHP files.

This would decouples the business logic from the layout.
## Todo

The following files contains form elements.
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Security/login.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Admin/Security/two_step_form.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Resetting/request.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Resetting/reset.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Profile/edit_profile_content.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Profile/edit_authentication_content.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Registration/register_content.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Resetting/request_content.html.twig
- [ ] https://github.com/sonata-project/SonataUserBundle/blob/3.x/Resources/views/Security/base_login.html.twig

jordisala1991:
## Subject

Closes #784 
Part of #1465 

## Changelog

```markdown
### Added
- Added `ResetPasswordRequestFormType` form type.
```


jordisala1991:
I have unified both the request of the password reset and the POST route. This way we can take advantage of the Symfony form validations and show errors to the user if he submits invalid data (no data in this case is the only considered invalid).

I have also added more functional tests to ensure when we submit invalid data on both forms we get proper response.

I don't like a lot having 10 dependencies for one controller but refactoring that requires more classes, so the idea is that we choose between:

1. Not using Symfony forms in the reset password request and have separate controllers (no validation, but whatever)
2. Use Symfony forms in the reset password request and have one controller (it can be refactored later, but not sure if I will invest more time to it)

wdyt  @VincentLanglet , is it worth?

VincentLanglet (approved):

VincentLanglet:
> Use Symfony forms in the reset password request and have one controller (it can be refactored later, but not sure if I will invest more time to it)

I'm ok with this so far

PR #1492, opened 2022-02-05
5.0.0-alpha-1

VincentLanglet:
@jordisala1991 I'm not sure if we should add the 6.x branch on dev-kit or if it's too early ?


jordisala1991 (approved):
We can do it after this release, dont worry

PR #1500, opened 2022-02-16
Improve user admin

jordisala1991:
## Subject

I am targeting this branch, because this is BC break.

Part of #1465 .

Maybe before it had sense to have multiple tabs because all the fields we had about social integrations, now that the admin is simplified IMO we should simplify the form show.

Also I changed the default form type for the roles. IMO we should deprecate the older one. Not sure if deprecate right away, because that would mean an unavoidable deprecation (because changing the form type is probably BC break for those doing integration tests on the form page) or removing directly here. Wdyt @VincentLanglet ?

**Before:**
image
image

**After:**
image

## Changelog

```markdown
### Fixed
- Simplified User admin to show all fields in a single page.
- Fixed User admin translation for group fields.

### Removed
- Removed `SecurityRolesType` to render roles on the User edit form.
```


jordisala1991:
I added the removal of the old type for rendering roles. I can move to another PR if needed.

VincentLanglet (dismissed):

VincentLanglet:
> I added the removal of the old type for rendering roles. I can move to another PR if needed.

Seems ok to me like this. But maybe we could add a line in the Changelog file

VincentLanglet (approved):

PR #1503, opened 2022-02-16
Make SonataAdminBundle dependency optional

### LINKED ISSUE #1499: Using this bundle without SonataAdminBundle
Hi.
Would making the SonataAdminBundle an optional dependency for v5 be something you could consider? It wouldn't be a BC break, since v5 is still in pre-release. And I could volunteer to implement the necessary changes.
In fact running it without the Admin bundle was possible with v4 with a bit of trickery in composer.json. But service configuration in v5 makes things a bit more complicated. The key changes required to make it work are:
1. Moving some config processing logic into this `if` block
```
# Sonata\UserBundle\DependencyInjection\SonataUserExtension.php
        if (isset($bundles['SonataAdminBundle'])) {
            $loader->load('admin.php');
            $loader->load(sprintf('admin_%s.php', $config['manager_type']));
        }
```
2. And `sonata.user.twig.global` (`Sonata\UserBundle\Twig\GlobalVariables`) shouldn't have a hard dependency on `Sonata\AdminBundle\Admin\Pool`.

Would you be open to such PR?

supersmile2009:
## Subject

This PR makes SonataAdminBundle dependency optional.

I am targeting 5.x branch, because v5 hasn't got to a final release yet and BC breaks can potentially be still accepted and there were no objections to that in the referenced issue.

Closes #1499

## Changelog

```markdown
### Changed
- SonataAdminBundle dependency is now optional. Trying to access `sonata_user.userAdmin` variable in Twig templates will throw a `\LogicException` if SonataAdminBundle is not installed.
```

## To do

- [x] Update the documentation;
- [x] Update the tests;
- [x] Add an upgrade note.

## Notes:
1. Do you think nullable admin Pool in `GlobalVariables` is OK or it would be better to conditionally register pool as another Twig variable (e. g. `sonata_user_admin`)?
2. I've checked the Media and Classification bundles (referenced in the linked issue as bundles with optional Admin dependency) and there seems to be no full-blown integration tests with running separate test-suits with and without Admin. So I took a modest approach with simply unit-testing the extension with a different config.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
         "doctrine/collections": "^1.6",
         "doctrine/common": "^3.1",
         "doctrine/persistence": "^2.1",
-        "sonata-project/admin-bundle": "^4.8",
         "sonata-project/doctrine-extensions": "^1.13",
         "sonata-project/form-extensions": "^1.4",
         "sonata-project/twig-extensions": "^1.3",
@@ -65,6 +64,7 @@
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.16",
         "psalm/plugin-symfony": "^3.0",
+        "sonata-project/admin-bundle": "^4.8",
         "sonata-project/block-bundle": "^4.0",
         "sonata-project/doctrine-orm-admin-bundle": "^4.0",
         "symfony/browser-kit": "^4.4 || ^5.3 || ^6.0",
@@ -78,7 +78,8 @@
         "sonata-project/doctrine-orm-admin-bundle": "<4.0"
     },
     "suggest": {
-        "sonata-project/doctrine-orm-admin-bundle": "For persisting entities"
+        "sonata-project/doctrine-orm-admin-bundle": "For persisting entities",
+        "sonata-project/admin-bundle": "For admin dashboard to manage users and other entities"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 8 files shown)


supersmile2009:
Failing tests are not caused by the changes in this PR. See https://github.com/sonata-project/SonataUserBundle/pull/1504

supersmile2009:
Disregard my last comment. It appears that some Symfony FrameworkBundle configs are cleverly applied depending on whether a certain dependency is expected to be available in non-dev install. I had to add `framework.assets` config to force activate AssetExtension for twig. Otherwise FrameworkBundle de-registers the service, because `symfony/asset` is not a prod dependency anymore (unlike it was before as a dependency of the admin bundle)

composer.json
```diff
@@ -65,6 +64,7 @@
         "phpunit/phpunit": "^9.5",
         "psalm/plugin-phpunit": "^0.16",
         "psalm/plugin-symfony": "^3.0",
+        "sonata-project/admin-bundle": "^4.8",
```
VincentLanglet, line 67: We need to add a line in the conflict 
```
"conflict": {
        "sonata-project/admin-bundle": "<4.8",
        "sonata-project/block-bundle": "<4.0",
        "sonata-project/doctrine-orm-admin-bundle": "<4.0"
    },
```

VincentLanglet (changes requested):

composer.json
supersmile2009: done

VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thanks @supersmile2009 !

PR #1504, opened 2022-02-16
Fix functional tests in v5

supersmile2009:
## Subject

I am targeting this branch, because it fixes failing tests on this branch.

Twig AssetExtension doesn't get registered when there's no `assets` entry in the `framework` config, which results in functional tests failing. Looks like older CI pipelines are green, so it must be related to some recent change in on of the dependencies.

## Changelog

```markdown
### Fixed
- Functional tests
```

### DIFF
--- a/tests/App/Resources/config/config_sf5.yaml
+++ b/tests/App/Resources/config/config_sf5.yaml
@@ -1,6 +1,7 @@
 framework:
     session:
         storage_factory_id: session.storage.factory.mock_file
+    assets:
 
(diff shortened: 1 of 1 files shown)


jordisala1991:
I cant make the test fail on 5.x, can you show a failing build?

supersmile2009:
Sorry, I accidentally used composer.lock from another branch to run the tests on 5.x locally. This fix belongs to the other PR.

PR #1505, opened 2022-02-17
Remove unstranslated translations

jordisala1991:
## Subject

Part of #1465


jordisala1991:
The translations file are a bit of madness... a lot of untranslated content. A lof of translations that are not even on english. Some translations needs to be removed, and we need to add the missing translations from FOSUser (already present on english)... it will take time to fix this and some of it might be done after the final release.

VincentLanglet (approved):

PR #1508, opened 2022-02-17
Fix show translations

jordisala1991:
## Subject

I am targeting this branch, because this is a bugfix.

Part of #1465 

## Changelog

```markdown
### Fixed
- Fixed show translations on User admin
```

### DIFF
--- a/src/Resources/translations/SonataUserBundle.bg.xliff
+++ b/src/Resources/translations/SonataUserBundle.bg.xliff
@@ -42,6 +42,10 @@
         <source>users</source>
         <target>Потребители</target>
       </trans-unit>
+      <trans-unit id="show.group_users">
+        <source>show.group_users</source>
+        <target>Потребители</target>
+      </trans-unit>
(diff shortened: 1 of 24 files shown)


VincentLanglet (approved):

src/Resources/translations/SonataUserBundle.bg.xliff
```diff
@@ -42,6 +42,10 @@
         <source>users</source>
         <target>Потребители</target>
       </trans-unit>
```
VincentLanglet, line 44: Shouldn't we remove the `users` translation ? which is now show.group_users

VincentLanglet (changes requested):

src/Resources/translations/SonataUserBundle.bg.xliff
jordisala1991: That translation is used on the side menu. So both translations are needed.

VincentLanglet (approved):

PR #1533, opened 2022-04-29
Fix UserAdmin breadcrumb name

aerrasti:
## Subject

The breadcrumb builder is using a classnameLabel that doesn't match the translation`breadcrumb.link_user_list`.

**Before**
image

**After**
image

I am targeting this branch, because the error is in the 5.x branch.

## Changelog

```markdown

### Changed
- Changed UserAdmin classnameLabel. This affects the breadcrumb translation generation. With this change Sonata will always pick the SonataUserBundle breadcrumbs translations by default.
```

### DIFF
--- a/src/Admin/Model/UserAdmin.php
+++ b/src/Admin/Model/UserAdmin.php
@@ -29,6 +29,7 @@ use Symfony\Component\Form\Extension\Core\Type\TextType;
 class UserAdmin extends AbstractAdmin
 {
     protected UserManagerInterface $userManager;
+    protected $classnameLabel = 'user';
 
     public function __construct(UserManagerInterface $userManager)
     {
--- a/src/Listener/UserListener.php
+++ b/src/Listener/UserListener.php
@@ -17,6 +17,7 @@ use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ODMClassMetadata;
 use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata as ORMClassMetadata;
 use Doctrine\Persistence\Event\LifecycleEventArgs;
 use Doctrine\Persistence\ObjectManager;
@@ -49,6 +50,9 @@ final class UserListener implements EventSubscriber
         ];
     }
 
+    /**
+     * @param LifecycleEventArgs<EntityManagerInterface|DocumentManager> $args
+     */
     public function prePersist(LifecycleEventArgs $args): void
     {
         $object = $args->getObject();
@@ -60,6 +64,9 @@ final class UserListener implements EventSubscriber
         $this->updateUser($object);
     }
 
+    /**
+     * @param LifecycleEventArgs<EntityManagerInterface|DocumentManager> $args
+     */
     public function preUpdate(LifecycleEventArgs $args): void
     {
         $object = $args->getObject();


jordisala1991:
I think we might want to investigate a bit. On Sonata 3 (With UserBundle 4) this problem didn't happened. Do you know if something changes between Sonata 3 and 4 that might be related @VincentLanglet ?

VincentLanglet:
> I think we might want to investigate a bit. On Sonata 3 (With UserBundle 4) this problem didn't happened. Do you know if something changes between Sonata 3 and 4 that might be related @VincentLanglet ?

Dunno.

But looking at the fix, it would meant that the classname label is not correctly set.
Currently, it's done this way.
```
final public function initialize(): void
    {
        if (null === $this->classnameLabel) {
            $namespaceSeparatorPos = strrpos($this->getClass(), '\\');
            $this->classnameLabel = false !== $namespaceSeparatorPos
                ? substr($this->getClass(), $namespaceSeparatorPos + 1)
                : $this->getClass();
        }

        $this->configure();

        foreach ($this->getExtensions() as $extension) {
            $extension->configure($this);
        }
    }
```
and getClass
```
final public function getClass(): string
    {
        if ($this->hasActiveSubClass()) {
            if ($this->hasParentFieldDescription()) {
                throw new \LogicException('Feature not implemented: an embedded admin cannot have subclass');
            }

            $subClass = $this->getRequest()->query->get('subclass');
            \assert(\is_string($subClass));

            if (!$this->hasSubClass($subClass)) {
                throw new \LogicException(sprintf('Subclass "%s" is not defined.', $subClass));
            }

            return $this->getSubClass($subClass);
        }

        // Do not use `$this->hasSubject()` and `$this->getSubject()` here to avoid infinite loop.
        // `getSubject` use `hasSubject()` which use `getObject()` which use `getClass()`.
        if (null !== $this->subject) {
            /** @phpstan-var class-string<T> $class */
            $class = ClassUtils::getClass($this->subject);

            return $class;
        }

        return $this->getModelClass();
    }
```

It would be interesting to dump the value to understand where is the issue.

The modelClass is set here
https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/Resources/config/admin_orm.php#L23

Coming from https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/DependencyInjection/SonataUserExtension.php#L90

@aerrasti What's the name of your user class ?
@jordisala1991 I think if the User class is not called `User` you'll end with this issue in the breadcrumb.

jordisala1991:
Hey I just checked and it happens the same with Sonata 3, so it's not something related to the upgrade.

VincentLanglet:
> Hey I just checked and it happens the same with Sonata 3, so it's not something related to the upgrade.

What's the user class you used ? Do I suppose correctly the origin of the bug ?

jordisala1991:
To me the problem is that, by default we add some translations for the breadcrumbs and all. But at the end, it depends if the class defined by the developer is called `User` or not. If it is called `SonataUser` or any other thing, the translation won't be used.

VincentLanglet:
> To me the problem is that, by default we add some translations for the breadcrumbs and all. But at the end, it depends if the class defined by the developer is called `User` or not. If it is called `SonataUser` or any other thing, the translation won't be used.

Yes, that's why the fix is correct.
But if someone used a custom translations breadcrumb.link_sonata_user_user_list, we will break his code with this patch.

Should it be a minor with a upgrade note "be careful" ?

jordisala1991:
Yes, we can consider it a minor. At worst, people will see their translations changed and used now the default of Sonata, which TBH is not that big of a deal, but we can give them the hint.

jordisala1991 (approved):

VincentLanglet (approved):

PR #1559, opened 2022-08-01
Include how to override Sonata User admin service in Documentation

Yippy:
## Subject

I am targeting this branch, because Sonata User documentation is missing a passage on how to override service just like Sonata Admin documentation,

## Changelog

```markdown
### Added
- Added example of overriding service in documentation
```

### DIFF
--- a/docs/reference/advanced_configuration.rst
+++ b/docs/reference/advanced_configuration.rst
@@ -1,6 +1,7 @@
 .. index::
     single: Advanced configuration
     single: Options
+    single: Override Service
 
 Advanced Configuration
 ======================
@@ -38,3 +39,30 @@ Full configuration options:
                 template: '@SonataUser/Admin/Security/Resetting/email.html.twig'
                 address: ~
                 sender_name: ~
+
+Override Service
+======================
+
+If you need to override the service of Sonata User admin, you can do it during the service declaration:
+
+.. code-block:: yaml
+
+    # config/services.yaml
+
+    services:
+        sonata.user.admin.user:
+            class: Sonata\UserBundle\Admin\Entity\UserAdmin
+            tags:
+                - name: sonata.admin
+                  model_class: '%sonata.user.user.class%'
+                  controller: '%sonata.user.admin.user.controller%'
+                  manager_type: orm
+                  group: sonata_user
+                  label: users
+                  translation_domain: SonataUserBundle
+                  label_translator_strategy: sonata.admin.label.strategy.underscore
+                  icon: '<i class=\'fa fa-users\'></i>'
+            arguments:
+                - '@sonata.user.manager.user'
+
+Please note that parameter ``%sonata.user.user.class%`` and ``%sonata.user.admin.user.controller`` refers to configuration options.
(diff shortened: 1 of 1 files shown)


VincentLanglet:
Do we have to repeat the tags ? 
Also this is following the standard way of overriding service from Symfony, so I'm not sure it requires extra-doc ; isn't it ?

Yippy:
> Do we have to repeat the tags ? Also this is following the standard way of overriding service from Symfony, so I'm not sure it requires extra-doc ; isn't it ?

That's true that it's a standard in overriding services, it was just an example to show the specific parameters to override sonata user services. Just like how Sonata Admin documentation shows examples of specific admin, figured it would save people time. If it's not needed then it's fine.

Hanmac:
but isn't it easier to configure the admin service over the admin->user->class configuration?

with that you can change the admin without touching the service?

Yippy:
> but isn't it easier to configure the admin service over the admin->user->class configuration?
> 
> with that you can change the admin without touching the service?

Are you referring to autowire, autoconfigure attributes? Though I did try using Dependency Injection attribute, doesn't seems to work for me properly.

In the end just passing a argument via through service was consistent, considering all my other admin are declared from services.yaml.

Hanmac:
> > but isn't it easier to configure the admin service over the admin->user->class configuration?
> > with that you can change the admin without touching the service?
> 
> Are you referring to autowire, autoconfigure attributes? Though I did try using Dependency Injection attribute, doesn't seems to work for me properly.
> 
> In the end just passing a argument via through service was consistent, considering all my other admin are declared from services.yaml.

i meant this config option:
https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/DependencyInjection/SonataUserExtension.php#L101

this sets the class for the `sonata.user.admin.user` service

should be
```yaml
sonata_user:
  admin:
    user:
      class: ClassNameThere
```

Yippy:
> > > but isn't it easier to configure the admin service over the admin->user->class configuration?
> > > with that you can change the admin without touching the service?
> > 
> > 
> > Are you referring to autowire, autoconfigure attributes? Though I did try using Dependency Injection attribute, doesn't seems to work for me properly.
> > In the end just passing a argument via through service was consistent, considering all my other admin are declared from services.yaml.
> 
> i meant this config option: https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/DependencyInjection/SonataUserExtension.php#L101
> 
> this sets the class for the `sonata.user.admin.user` service
> 
> should be
> 
> ```yaml
> sonata_user:
>   admin:
>     user:
>       class: ClassNameThere
> ```

Yeah I know about class configuration option, it's on the same page as that document. I don't think there is a option for passing services into the configuration option. For example I need to send a service into the Admin, such as @security.token_storage for the admin to display a custom view if the user is itself.

Hanmac:
have you tried to use setter autowiring?

https://symfony.com/doc/current/service_container/calls.html
https://symfony.com/doc/current/service_container/autowiring.html#autowiring-calls

Yippy:
> have you tried to use setter autowiring?
> 
> https://symfony.com/doc/current/service_container/calls.html https://symfony.com/doc/current/service_container/autowiring.html#autowiring-calls

It's simpler to just override the Sonata User admin service, to include missing services.
For controllers I just amend getSubscribedServices(), to include services.

The documentation of amending Sonata User admin service, was missing. Just figured that people may want to know how to do it, just like Sonata Admin documentation explains how to amend admin services. But this PR is specifically for Sonata User bundle... hence the PR.

PR #1562, opened 2022-08-11
fix: add UniqueEntity constraint to email field.

m-ar-c:
## Subject

This PR add `UniqueEntity` constraint to email field in [validation.xml](https://github.com/sonata-project/SonataUserBundle/blob/5.x/src/Resources/config/validation.xml).

Why :
The email property is defined as unique in the DB, so we need to
constraint the field to unique in the form in order to prevent a
`UniqueConstraintViolationException` exception.

I am targeting this branch, because it is BC.

## Changelog
```markdown
### Fixed
- Add UniqueEntity constraint to email field in order to prevent UniqueConstraintViolationException exception.
```

### DIFF
--- a/src/Resources/config/validation.xml
+++ b/src/Resources/config/validation.xml
@@ -3,7 +3,7 @@
     <class name="Sonata\UserBundle\Model\User">
         <constraint name="Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity">
             <option name="fields">email</option>
-	    <option name="groups">
+            <option name="groups">
                 <value>Registration</value>
                 <value>Profile</value>
             </option>
--- a/src/Resources/config/validation.xml
+++ b/src/Resources/config/validation.xml
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <constraint-mapping xmlns="http://symfony.com/schema/dic/constraint-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/constraint-mapping http://symfony.com/schema/dic/constraint-mapping/constraint-mapping-1.0.xsd">
     <class name="Sonata\UserBundle\Model\User">
+        <constraint name="Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity">
+            <option name="fields">email</option>
+        </constraint>
         <property name="username">
             <constraint name="NotBlank">
                 <option name="groups">
@@ -18,6 +21,7 @@
             </constraint>
         </property>
         <property name="email">
+	    <constraint name="Email"/>
             <constraint name="NotBlank">
                 <option name="groups">
                     <value>Registration</value>
--- a/src/Resources/config/validation.xml
+++ b/src/Resources/config/validation.xml
@@ -21,7 +21,6 @@
             </constraint>
         </property>
         <property name="email">
-	    <constraint name="Email"/>
             <constraint name="NotBlank">
                 <option name="groups">
                     <value>Registration</value>
--- a/src/Resources/config/validation.xml
+++ b/src/Resources/config/validation.xml
@@ -3,6 +3,10 @@
     <class name="Sonata\UserBundle\Model\User">
         <constraint name="Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity">
             <option name="fields">email</option>
+	    <option name="groups">
+                <value>Registration</value>
+                <value>Profile</value>
+            </option>
         </constraint>
         <property name="username">
             <constraint name="NotBlank">


Hanmac:
i see this a bit critical, for one of my projects i need to have users with different ids but with same email

it is easy to overwrite the doctrine setting for uniq column, but it isn't possible to remove unique entry validation.
that only works if you can somehow set a different valdiation group (which i don't know if it is possible with sonata user bundle)

VincentLanglet:
> i see this a bit critical, for one of my projects i need to have users with different ids but with same email
> 
> it is easy to overwrite the doctrine setting for uniq column, but it isn't possible to remove unique entry validation. that only works if you can somehow set a different valdiation group (which i don't know if it is possible with sonata user bundle)

As long as we define the property as unique in database, we should be consistent and add an annotation IMHO.

How do you override the unique definition ? It could be interesting to see if you cannot override the annotation too.
Because it's kinda the job of the one overriding the `unique=true` to remove the annotation and not the bundle to not provide an annotation in order to allow people to override. (cc @jordisala1991 if you know how)

Hanmac:
For ORM, `@AttributeOverride`
* https://symfony.com/doc/current/bundles/override.html#entities-entity-mapping
* https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/inheritance-mapping.html#attribute-override

for Validator:
* https://symfony.com/doc/current/bundles/override.html#validation-metadata
> To overcome this, the 3rd party bundle needs to have configuration for [validation groups](https://symfony.com/doc/current/validation/groups.html). For instance, the FOSUserBundle has this configuration. To create your own validation, add the constraints to a new validation group

the other contraints have groups like "Registration", the new one has not.

Also it should probably be better configured where the Validation Groups are used.

VincentLanglet:
Indeed since there is validation group everywhere we should use it too.

Does adding 
```
                <option name="groups">
                    <value>Registration</value>
                    <value>Profile</value>
                </option>
```
would solve your issue @Hanmac ?

Hanmac:
> Indeed since there is validation group everywhere we should use it too.
> 
> Does adding
> 
> ```
>                 <option name="groups">
>                     <value>Registration</value>
>                     <value>Profile</value>
>                 </option>
> ```
> 
> would solve your issue @Hanmac ?

i would need to debug it, but i could work with that i think

VincentLanglet:
If you have time to try in order to confirm it, it would be great. This would avoid us to block you ^^

m-ar-c:
So, to add validation group, I just need to remove the line (24) `<constraint name="Email"/>` I added, since there already was a
```xml
            <constraint name="Email">
                <option name="groups">
                    <value>Registration</value>
                    <value>Profile</value>
                </option>
            </constraint>
```
block in the code. Am I right ? (I tested of course, unique validation on email still works with the line removed).

src/Resources/config/validation.xml
```diff
@@ -18,6 +21,7 @@
             </constraint>
         </property>
         <property name="email">
+	    <constraint name="Email"/>
```
jordisala1991: This line could be removed.

src/Resources/config/validation.xml
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <constraint-mapping xmlns="http://symfony.com/schema/dic/constraint-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/constraint-mapping http://symfony.com/schema/dic/constraint-mapping/constraint-mapping-1.0.xsd">
     <class name="Sonata\UserBundle\Model\User">
+        <constraint name="Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity">
+            <option name="fields">email</option>
```
jordisala1991, line 5:
```suggestion
            <option name="fields">email</option>
            <option name="groups">
                <value>Registration</value>
                <value>Profile</value>
            </option>
```

src/Resources/config/validation.xml
```diff
@@ -1,6 +1,13 @@
...
+	    <option name="groups">
```
VincentLanglet, line 6:
```suggestion
	        <option name="groups">
```
I think the indentation is not perfect

VincentLanglet (changes requested):

src/Resources/config/validation.xml
VincentLanglet:
```suggestion
	        <option name="groups">
```
I'm not sure but the indent seems weird

VincentLanglet (changes requested):

VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thanks @m-ar-c 

@Hanmac You can override the validation groups for the admin, take a look at the model user admin.

PR #1576, opened 2022-10-13
Make Reset Password Optional

### LINKED ISSUE #1491: Make reset password optional
## Feature Request

It would be nice to enable / disable reset password and avoid dependencies with mailer.

Extracted from: #1465

Hanmac:
## Subject

I am targeting this branch, because its BC break (most likely).

Closes #1491.

## Changelog

```markdown
### Changed
- Make Reset Password Optional
```

## To do

- [x] Update the tests, i probably need new App Configs without Mailer
- [ ] Update the documentation;
- [ ] Add an upgrade note.


Hanmac:
@VincentLanglet should i try to fix this errors even if i didn't touched this code?

Hanmac:
`getFlashBag` needs https://github.com/symfony/symfony/pull/46491 https://github.com/symfony/symfony/commit/026c74820877d35b24fa3996aa05d4f10f6fce4f
which only exists in Symfony 6.2+
so just do a `$session instance of Session` around it?

VincentLanglet:
I made https://github.com/sonata-project/SonataUserBundle/pull/1595 ; you just need to rebase once 5.x will be in 6.x.

Btw, the 6.x release won't be soon. Can't you do the PR on the 5.x ? I don't see a real BC-break

src/Action/LoginAction.php
```diff
@@ -52,7 +54,8 @@ public function __construct(
         TemplateRegistryInterface $templateRegistry,
         TokenStorageInterface $tokenStorage,
         TranslatorInterface $translator,
-        ?CsrfTokenManagerInterface $csrfTokenManager = null
+        ?CsrfTokenManagerInterface $csrfTokenManager = null,
+        bool $resetMail = false
```
VincentLanglet: To be BC we should set the default value to `true`. (In case someone does `new LoginAction(...)`)

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -69,11 +69,27 @@ public function load(array $configs, ContainerBuilder $container): void
...
         $this->configureClass($config, $container);
-        $this->configureMailer($config, $container);
         $this->configureDefaultAvatar($config['profile'], $container);
         if (isset($bundles['SonataAdminBundle'])) {
             $this->configureAdmin($config['admin'], $container);
+            // check for reset configuration
+            if (isset($config['resetting']['email']['address'], $config['resetting']['email']['sender_name'])) {
+                $loader->load('actions_resetting.php');
+                $mailerNeeded = true;
+                $resetting = true;
```
VincentLanglet: Do we really need two variables ? We could do
```
if ($resetting) {            
$loader->load('mailer.php');
            $this->configureMailer($config, $container);
            $this->configureResetting($config['resetting'], $container);
        }
```

src/Action/LoginAction.php
Hanmac: should the default be updated in the `actions.php` too?

VincentLanglet: yes indeed

Hanmac: i made `resetMail` default to true

tests/Action/LoginActionTest.php
```diff
@@ -228,7 +228,8 @@ private function getAction(): LoginAction
             $this->templateRegistry,
             $this->tokenStorage,
             $this->translator,
-            $this->csrfTokenManager
+            $this->csrfTokenManager,
+            true,
```
Hanmac: @VincentLanglet if i want to make a mock test case for resetting false, should i make a whole new LoginAction(withoutReset)Test class or should i try to extend this one?

VincentLanglet: Extending doesn't seem a bad idea in this case if it's possible

Hanmac: i extended the Test Data Provider with Resetting=false

VincentLanglet (commented):
Seems ok ; but didn't you plan to write doc and update note ?

Hanmac:
> Seems ok ; but didn't you plan to write doc and update note ?

I'm not very good at documentation and update note stuff

i probably need to write something like this:
> When updating, please require "symfony/mailer" and "symfony/mime" manually
> Or to disable the function, remove the configuration

IMO isn't this still a BC break? and shouldn't that be marked for remove next-major?

composer.json
```diff
@@ -35,8 +35,6 @@
         "symfony/framework-bundle": "^4.4 || ^5.4 || ^6.0",
         "symfony/http-foundation": "^4.4 || ^5.4 || ^6.0",
         "symfony/http-kernel": "^4.4 || ^5.4 || ^6.0",
-        "symfony/mailer": "^4.4 || ^5.4 || ^6.0",
-        "symfony/mime": "^4.4.10 || ^5.4 || ^6.0",
```
Hanmac: @VincentLanglet i'm unsure, is this a BC Break?

updating with composer would remove the dependency and composer would remove the packages?
which then causes a problem that the current default config (and what the user might have configured)
are still looking for the mailer packages? (i mean it probably would cause problems with cache:clear after composer update, so the user should see the problem directly?)

VincentLanglet: Friendly ping @jordisala1991.
I don't remember where but I think we already did such change (which an upgrade note).

The other option would be to keep those dependencies and only remove them in 6.x branch

composer.json
```diff
@@ -70,9 +68,10 @@
         "sonata-project/block-bundle": "^4.11",
         "sonata-project/doctrine-orm-admin-bundle": "^4.0",
         "symfony/browser-kit": "^4.4 || ^5.4 || ^6.0",
-        "symfony/console": "^4.4 || ^5.4 || ^6.0",
```
jordisala1991: Why removing this dep?

composer.json
jordisala1991: Since you have to configure mailer, it should be a dep you have when installing this bundle. IMO we can try moving it to require-dev

src/DependencyInjection/SonataUserExtension.php
```diff
@@ -69,12 +69,23 @@ public function load(array $configs, ContainerBuilder $container): void
...
+        $resetting = false;
+
         $this->configureClass($config, $container);
-        $this->configureMailer($config, $container);
         $this->configureDefaultAvatar($config['profile'], $container);
         if (isset($bundles['SonataAdminBundle'])) {
             $this->configureAdmin($config['admin'], $container);
-            $this->configureResetting($config['resetting'], $container);
+            // check for reset configuration
+            if (isset($config['resetting']['email']['address'], $config['resetting']['email']['sender_name'])) {
```
jordisala1991, line 80: IMO this is missing a check to really validate the user has required deps (since they will be optional on require-dev) wdyt ?

src/DependencyInjection/Configuration.php
```diff
@@ -113,8 +112,8 @@ private function addResettingSection(ArrayNodeDefinition $node): void
                             ->addDefaultsIfNotSet()
                             ->children()
                                 ->scalarNode('template')->cannotBeEmpty()->defaultValue('@SonataUser/Admin/Security/Resetting/email.html.twig')->end()
-                                ->scalarNode('address')->isRequired()->cannotBeEmpty()->end()
```
jordisala1991, line 116: the cannotBeEmpty should be kept, otherwise it allows defining the key with a empty value, right? like the 'template' key.

jordisala1991:
WDYT about adding an option to disable reset mail (default to enabled) and deprecate not defining a value for it (like symfony did for the enable_authenticator_manager). Not sure how they did it or how hard is to implement. But would make the feature easier to discover.

Also, some docs about how to disable/enable the feature, would be nice.

VincentLanglet:
> WDYT about adding an option to disable reset mail (default to enabled) and deprecate not defining a value for it (like symfony did for the enable_authenticator_manager). Not sure how they did it or how hard is to implement. But would make the feature easier to discover.

There is already 
```
$config['resetting']['email']['address']
$config['resetting']['email']['sender_name']
```
which are two values required for reset mail

If you don't want reset mail, you remove those value ; and if you want reset mail, you set those values.
Isn't it enough ? It avoid handling case when you enable the reset mail but doesn't set those values...

composer.json
Hanmac: huch i didn't do it, probably composer did because it is already a required dependency

src/DependencyInjection/Configuration.php
```diff
@@ -112,8 +112,8 @@ private function addResettingSection(ArrayNodeDefinition $node): void
                             ->addDefaultsIfNotSet()
                             ->children()
                                 ->scalarNode('template')->cannotBeEmpty()->defaultValue('@SonataUser/Admin/Security/Resetting/email.html.twig')->end()
-                                ->scalarNode('address')->end()
-                                ->scalarNode('sender_name')->end()
+                                ->scalarNode('address')->cannotBeEmpty()->end()
+                                ->scalarNode('sender_name')->cannotBeEmpty()->end()
```
VincentLanglet, line 116: This was better without the cannotBeEmpty no ? This way it's really optional.
What do you plan to do @Hanmac

Hanmac: i removed the required on the top of resetting,
so when you just don't configure it, it should not require the mailer

the main goal for this MR, outside from making it not require the mailer if not need to,
is for a very fresh symfony project to just `composer require sonata-project/user-bundle` without having it crash in the `cache:clear` because of the missing sonata_user.yml having missing the resetting configuration

VincentLanglet: Yes, and I agree with this change.
That's why IMHO we shouldn't revert this.

The goal of this PR is to remove the `cannotBeEmpty`, so I was surprised you added it back in the last commit.

jordisala1991: Cannot be empty does not make the field required, it just make it when you define it, you cannot define it empty.

You are talking about isRequired.

VincentLanglet: Oh I see ; my bad.

VincentLanglet:
Except the need of a rebase, this PR need to update the doc in order to explain how to enable/disable this feature @Hanmac

Hanmac:
For the Documentation i need to think about what the best way to write it

VincentLanglet:
This part need to be updated https://github.com/sonata-project/SonataUserBundle/blob/5.x/docs/reference/installation.rst#configuration
since 
```
resetting:
        email:
            address: sonata@localhost
            sender_name: Sonata Admin
```
is not mandatory.

I would say the routing too https://github.com/sonata-project/SonataUserBundle/blob/5.x/docs/reference/installation.rst#integrating-the-bundle-into-the-sonata-admin-bundle

Maybe this section too https://github.com/sonata-project/SonataUserBundle/blob/5.x/docs/reference/installation.rst#mailer-configuration

Basically we need for every reset-specific code to give an exemple with and without the feature.

VincentLanglet:
> For the Documentation i need to think about what the best way to write it

Hi @Hanmac, how are you doing with the doc update ?
This PR would be a nice improvement to add to this bundle :)

VincentLanglet:
Friendly ping @Hanmac ; any time for the doc update ? :)

Hanmac:
I still need to do this an some Doc Update 

Maybe next Weekend

VincentLanglet:
> I still need to do this an some Doc Update
> 
> Maybe next Weekend

Thanks.

You'll also need a rebase.

VincentLanglet (dismissed):

VincentLanglet:
Let's forget the doc, if you rebase @Hanmac I'll merge it

PR #1603, opened 2023-03-13
Allow doctrine annotations 2.0

jordisala1991:
## Subject

I am targeting this branch, because this only affects php-cs-fixer.


jordisala1991:
Doctrine annotations needs to be removed when we migrate to attributes on tests, but I wont be able to do it soon, so I allow 2.0 to not block dev kit

VincentLanglet (approved):

PR #1651, opened 2023-06-13
UserListener: use event_listener instead of subscriber

### LINKED ISSUE #1650: Deprecation: UserListener needs to be done as listener instead of subscriber
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
# Put the result here.
sonata-project/user-bundle 5.9.0
```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
# Put the result here.
symfony/doctrine-bridge v6.3.0
```

</p>
</details>

#### PHP version

```
$ php -v
# Put the result here.
PHP 8.2.6
```

## Steps to reproduce

Open a Page in Backend and use Profiler to see Deprecations

## Expected results

No Deprecations

## Actual results

> 'symfony/doctrine-bridge', '6.3', 'Using Doctrine subscribers as services is deprecated, declare listeners instead'

Because of https://github.com/symfony/symfony/pull/49918 and https://github.com/doctrine/DoctrineBundle/pull/1664,
`doctrine.event_subscriber` is deprecated and should use listener instead

Should that be done as https://symfony.com/doc/current/doctrine/events.html#doctrine-entity-listeners directly instead of listening to all entities? technically it should not look for `Sonata\UserBundle\Model\UserInterface` but for `sonata.user.user.class` ?

for a possible MR, should i still use the php config, or should i use [#[AsEntityListener]](https://github.com/doctrine/DoctrineBundle/blob/2.8.x/Attribute/AsEntityListener.php) ?

Hanmac:
## Subject

I am targeting this branch, because it should be without BC break?

Closes #1650.

## Changelog

```markdown
### Fixed
- Deprecation of Event Subscribers on Symfony 6.3. The UserListener now uses Event Listeners
```

From Coverage testing, it seems that current phpunit test doesn't check the same for ODM too.
Do we need more Unit tests, that explicit test ODM?

* `doctrine.orm.entity_listener` would be more specific for ORM, but it seems something like that doesn't exist for ODM.

### DIFF
--- a/src/Resources/config/mongodb.php
+++ b/src/Resources/config/mongodb.php
@@ -28,7 +28,12 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             ])
 
         ->set('sonata.user.listener.user', UserListener::class)
-            ->tag('doctrine_mongodb.odm.event_subscriber')
+            ->tag('doctrine.event_listener', [
+                'event' => 'prePersist'
+            ])
+            ->tag('doctrine.event_listener', [
+                'event' => 'preUpdate'
+            ])
             ->args([
                 service('sonata.user.util.canonical_fields_updater'),
                 service('sonata.user.manager.user'),
--- a/src/Resources/config/mongodb.php
+++ b/src/Resources/config/mongodb.php
@@ -28,11 +28,11 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             ])
 
         ->set('sonata.user.listener.user', UserListener::class)
-            ->tag('doctrine.event_listener', [
-                'event' => 'prePersist'
+            ->tag('doctrine_mongodb.odm.event_listener', [
+                'event' => 'prePersist',
             ])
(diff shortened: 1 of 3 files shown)


src/Resources/config/mongodb.php
```diff
@@ -28,7 +28,12 @@
             ])
 
         ->set('sonata.user.listener.user', UserListener::class)
-            ->tag('doctrine_mongodb.odm.event_subscriber')
+            ->tag('doctrine.event_listener', [
```
jordisala1991: make sure the mongodb tag is like this please, with the subscriber it was different.

Hanmac: you are right, seems there was some faulty documentation yesterday

VincentLanglet (approved):

jordisala1991:
Thanks @Hanmac

PR #1661, opened 2023-09-30
Use multiple options by default for `RolesMatrixType` form

phansys:
## Subject

Use multiple options by default for `RolesMatrixType` form

I am targeting this branch, because this change seems to follow the natural use case for this form, where multiple options for the role matrix are allowed (rendered as checkboxes instead of radio buttons):
Roles Matrix

## Changelog

```markdown
### Added
- Default value `true` for the "multiple" option at `RolesMatrixType` form type

### Changed
- `RolesMatrixType` form type allows multiple values by default
```


VincentLanglet (approved):

VincentLanglet:
Thanks

PR #1662, opened 2023-10-21
Enable autowire and autoconfigure for sonata.user.admin.user service

dmikhinev:
## Subject

Enable autowire and autoconfigure for sonata.user.admin.user service

I am targeting this branch, because changes are related to current version.

## Changelog

```markdown
### Added
- Enabled autowire and autoconfigure for sonata.user.admin.user service

```


jordisala1991:
Why should we enable autowire and autoconfiguration for this service if it already injects all dependencies and tags? I think this change is not needed.

dmikhinev:
@jordisala1991 what if i need to inject some additional services? For example file uploader service to upload user profile photo?

VincentLanglet:
> @jordisala1991 what if i need to inject some additional services? For example file uploader service to upload user profile photo?

Then it's your service which will need autoconfigure, not the one defined by sonata no ?

Anyway, since none of our service use those tags, I think we should keep consistency.

PR #1669, opened 2023-12-19
Allow Symfony 7

jordisala1991:
## Subject

## Changelog

```markdown
### Added
- Add support for Symfony 7.
```


src/Security/Authorization/Voter/UserAclVoter.php
```diff
@@ -14,10 +14,10 @@
 namespace Sonata\UserBundle\Security\Authorization\Voter;
 
 use Sonata\UserBundle\Model\UserInterface;
-use Symfony\Component\Security\Acl\Voter\AclVoter;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Authorization\Voter\VoterInterface;
 
-final class UserAclVoter extends AclVoter
+final class UserAclVoter implements VoterInterface
```
jordisala1991: I am not sure why we needed to extend the acl class if this is not using anything from it, do you know @VincentLanglet ?

The acl package is not really compatible with sf 7.0, the aclvoter misses some typehints

Doing this change is bc break. What should we do?

VincentLanglet: I have no idea why it was done like this...

But the security-acl was supposed to be SF-7 compatible https://github.com/symfony/security-acl/blob/main/composer.json ; is there somethng to report to symfony ?

jordisala1991: Yep, this interface is not ready for SF 7: https://github.com/symfony/security-acl/blob/main/Voter/AclVoter.php, there are no typehints.

VincentLanglet: Indeed we need https://github.com/symfony/security-acl/pull/112

jordisala1991: The security-acl package is only used here, for an extends that does not really make sense. Unless I am missing something, the only BC break would be if someone tries to instantiate this class (since it is final) or redeclares the dependency injection... but this class is not expected to be construct by anyone else, since it is hook on the security system using the voter tag.

We could remove dep here and implement normal voter interface. The BC break is not really big

If someone uses security-acl, they need to add the bundle, and the component, and configure them on the security yaml.

VincentLanglet: There is some ACL-related doc in this bundle like https://github.com/sonata-project/SonataUserBundle/blob/5.x/docs/reference/installation.rst#acl-configuration

I dunno if we extends the AclVoter to decorate it and replacing it in the service injection. :(
But I never used this bundle...

If we don't need ACL here, let's remove it.

jordisala1991: security.acl.voter.user_permissions 

It could be replacing the service indeed, but AFAIK the service is only declared on the bundle: https://github.com/symfony/acl-bundle/blob/main/src/Resources/config/acl.xml#L25C9-L25C114

So it is not really replacing it, but declaring another one, the name is different.

VincentLanglet: Let's try to remove the extends then

core23: I've created a PR on the securty-acl bundle that is only waiting to be merged
https://github.com/symfony/security-acl/pull/116

JE4GLE:
Please re-open this PR

VincentLanglet:
> Please re-open this PR

As you can see the author is not working on it. Feel free to cherry pick the work and open your own PR

jordisala1991:
Now, I think we can recover this PR

src/Security/Authorization/Voter/UserAclVoter.php
jordisala1991: Fixed on security-acl

jordisala1991:
Related dev-kit change: https://github.com/sonata-project/dev-kit/pull/2842

VincentLanglet (approved):

PR #1677, opened 2024-02-27
Compatibility with doctrine dbal 4

RobinDev:
## Compatibility with doctrine dbal 4

Just switching the deprecated array type to json - cf [Upgrade Guide](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-removed-array-and-object-column-types) 
 
I am targeting this branch, because it's a patch wich is not introducing bc break.

About BC Break, i am not sure, because it's required to update from unserialized to json encoded.
  
## Changelog
 
```markdown 
### Added
- BaseUser3 with roles mapped type as `json` to be compatible with ORM 3
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "dama/doctrine-test-bundle": "^7.0",
         "doctrine/doctrine-bundle": "^2.7",
         "doctrine/mongodb-odm": "^2.3",
-        "doctrine/orm": "^2.14",
+        "doctrine/orm": "^2.13 || ^3.0",
         "egulias/email-validator": "^3.1 || ^4.0",
         "friendsofphp/php-cs-fixer": "^3.4",
         "matthiasnoback/symfony-config-test": "^4.2",
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "dama/doctrine-test-bundle": "^7.0",
         "doctrine/doctrine-bundle": "^2.7",
         "doctrine/mongodb-odm": "^2.3",
-        "doctrine/orm": "^2.13 || ^3.0",
+        "doctrine/orm": "^2.14 || ^3.0",
         "egulias/email-validator": "^3.1 || ^4.0",
         "friendsofphp/php-cs-fixer": "^3.4",
         "matthiasnoback/symfony-config-test": "^4.2",
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -82,6 +82,7 @@ And then create the corresponding entity, ``src/Entity/SonataUserUser``::
     use Doctrine\DBAL\Types\Types;
     use Doctrine\ORM\Mapping as ORM;
     use Sonata\UserBundle\Entity\BaseUser;
+    // or `Sonata\UserBundle\Entity\BaseUser3` as BaseUser if you upgrade to doctrine/orm ^3
 
     #[ORM\Entity]
     #[ORM\Table(name: 'user__user')]
--- a/src/Entity/BaseUser3.php
+++ b/src/Entity/BaseUser3.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\UserBundle\Entity;
+
+use Sonata\UserBundle\Model\User as AbstractedUser;
+
+class BaseUser3 extends AbstractedUser
+{
+    public function prePersist(): void
+    {
+        $this->createdAt = new \DateTime();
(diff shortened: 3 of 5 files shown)


RobinDev:
Not ready, i forget to upgrade dependencies...

RobinDev:
> About BC Break, i am not sure, because it's required to update from unserialized to json encoded.

I run a migration on a current project using sonataUserBundle, should I add it to this PR ?

```php
<?php

declare(strict_types=1);

namespace DoctrineMigrations;

use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

final class Version1709056840 extends AbstractMigration
{
    public function up(Schema $schema): void
    { 
        $rows = $this->connection->fetchAllAssociative('SELECT id,roles FROM user');

        /** @var array{'id': scalar, 'roles': string}[] $rows */
        foreach ($rows as $row) {
            $id = $row['id'];
            $roles = json_encode(unserialize($row['roles']));
            $this->connection->executeQuery('UPDATE user SET roles = :roles WHERE id = :id', ['roles' => $roles, 'id' => $id]);
        }
    }
}
```

VincentLanglet:
WDYT @jordisala1991 ? Should we consider this as a BC break ?

core23:
I would handle this as a BC break, because without a migration you can't load the roles anymore.

As this bundle does not provide any auto-migration script, there is no way to run `composer update` on a project without any problems.

VincentLanglet:
> I would handle this as a BC break, because without a migration you can't load the roles anymore.
> 
> As this bundle does not provide any auto-migration script, there is no way to run `composer update` on a project without any problems.

Yeah I was afraid about this.

Maybe we could have two files `BaseUser.orm2.xml` and `BaseUser.orm3.xml` and one of the two solution
- Automatically load the right file based on the ORM version
- Adding a configuration option to tell which file should be loaded

Wdyt @RobinDev ?

RobinDev:
I think your suggestions permit to ship it now and maintain compatibility with dbal 3 for a moment... but it's adding some complexity to maintain.

I would have a preference to target the 6.X branch wich is available via composer.

It will permit for the projects that required an upgrade of dbal to target the next version.

Feel free to modify my PR, i will not be available during the next 8 days.

RobinDev:
I made an other PR on 6.X #1680

VincentLanglet:
> I think your suggestions permit to ship it now and maintain compatibility with dbal 3 for a moment... but it's adding some complexity to maintain.
> 
> I would have a preference to target the 6.X branch wich is available via composer.

There is less complexity maintaining one branch, even with some config loaded conditionally rather than maintaining two branch. We don't plan to release 6.x branch soon, so I would say we don't plan to merge any PR on this branch.

So we have to find another solution...

VincentLanglet:
Maybe you have a suggestion @jordisala1991 @dmaicher ?

RobinDev:
I follow your suggestions.

So if an user want to use doctrine/orm ^3 : 

* the user must extend is entity with `BaseUser3` instead of `BaseUser` 
* the user must manage the migration if the entity was existing (sonata/user-bundle upgrade)

VincentLanglet:
> I follow your suggestions.
> 
> So if an user want to use doctrine/orm ^3 :
> 
> * the user must extend is entity with `BaseUser3` instead of `BaseUser`
> * the user must manage the migration if the entity was existing (sonata/user-bundle upgrade)

Seems like a nice solution without bc break. WDYT @core23 ?

VincentLanglet:
I fixed the cs, you can rebase https://github.com/sonata-project/SonataUserBundle/pull/1681

RobinDev:
Done

VincentLanglet:
> Done

Thanks, I'll try to get some opinions from others reviewers.

Ping @phansys @core23 @jordisala1991 @dmaicher @franmomu wdyt ?

VincentLanglet:
Thanks @RobinDev

rande:
Hello, coming late to this change. 

All this code can be updated to use a Doctrine Listener to change the mapping type without creating a new class and definition, this will make the migration path way easier.

This is why we have the https://github.com/sonata-project/sonata-doctrine-extensions/blob/2.x/src/Mapper/DoctrineCollector.php in the `sonata-project/doctrine-extensions` project

rande:
Ping @RobinDev @VincentLanglet - Do you want me to do a MR on this to avoid introducing a BC break ?

VincentLanglet:
> Ping @RobinDev @VincentLanglet - Do you want me to do a MR on this to avoid introducing a BC break ?

Sure !

rande:
Done: https://github.com/sonata-project/SonataUserBundle/pull/1685

Hanmac:
Are we sure about that array type?

https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#array says:
> This type is deprecated since 3.4.0, use [json](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#json) instead.

right now, I get:

> Unknown column type "array" requested.

Also we might need some better way to migrate the roles data?

PR #1680, opened 2024-03-20
Compatibility with doctrine dbal 4 

RobinDev:
## Compatibility with doctrine dbal 4

Just switching the deprecated array type to json - cf [Upgrade Guide](https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md#bc-break-removed-array-and-object-column-types) 
 
I am targeting this branch, because change cannot be done without a BC-break. 
  
## Changelog
 
```markdown 
### Changed
- roles mapped type (orm) from `array` to `json`
  
``` 

## Upgrade

A migration must be done, example : 

```php
<?php

declare(strict_types=1);

namespace DoctrineMigrations;

use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

final class Version1709056840 extends AbstractMigration
{
    public function up(Schema $schema): void
    { 
        $rows = $this->connection->fetchAllAssociative('SELECT id,roles FROM user');

        /** @var array{'id': scalar, 'roles': string}[] $rows */
        foreach ($rows as $row) {
            $id = $row['id'];
            $roles = json_encode(unserialize($row['roles']));
            $this->connection->executeQuery('UPDATE user SET roles = :roles WHERE id = :id', ['roles' => $roles, 'id' => $id]);
        }
    }
}
```


VincentLanglet:
See https://github.com/sonata-project/SonataUserBundle/pull/1677#issuecomment-2010488298

Let's wait for https://github.com/doctrine/orm/issues/11388

PR #1685, opened 2024-05-15
Make the code compatible with DBAL 3 and 4 by swithing the  type when…

rande:
## Make the code compatible with DBAL 3 and 4 by swithing the type when the Mapping is loaded

I am targeting this branch, because because it introduces a non necessary BC break. For reference: https://github.com/sonata-project/SonataUserBundle/pull/1677

## Changelog
 
```markdown 
### Removed
- BaseUser3 doctrine definition

### Added
- Mapping listener to make the same entity works with DBAL 2, 3 or 4
``` 

## To do

- [x] Review PR;


src/Entity/BaseUser3.php
```diff
@@ -13,18 +13,10 @@
...
 namespace Sonata\UserBundle\Entity;
 
-use Sonata\UserBundle\Model\User as AbstractedUser;
-
-class BaseUser3 extends AbstractedUser
+/**
+ * This class is used to avoid error for users that have installed the version
+ * 5.12.0 of the SonataUserBundle and have not updated their User class.
+ */
+class BaseUser3 extends BaseUser
```
VincentLanglet, line 20: The [BaseUser3.orm.xml](https://github.com/sonata-project/SonataUserBundle/pull/1685/files#diff-04c482ac96eadf5ce39b5f6dd08d04d6694842f96c5295076c94774a10add5c8) is not needed for this class ?

VincentLanglet:
There is the error
```
In MappingException.php line 112:
                                                                               
  The column type of attribute 'roles' on class 'Sonata\UserBundle\Tests\App\  
  Entity\User' could not be changed. 
```

src/Entity/BaseUser3.php
rande: Not anymore the Mapping listener is changing the type on during the ClassMetadata loading process. `array` type is working until DBAL 3 included, DBAL 4 removes the class ArrayType class.

VincentLanglet: And the BaseUser definition will be used for BaseUser3 because the class extends the BaseUser class ?

rande: yes

rande:
@VincentLanglet hopefully this is good to go!

src/Listener/DoctrineMappingListener.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        if (!$metadata->hasField('roles')) {
+            return;
+        }
+
+        /**
+         * @psalm-suppress InvalidPropertyAssignmentValue
+         */
+        $metadata->fieldMappings['roles']['type'] = 'array';
```
VincentLanglet, line 52: Is it the right way to change the value ?

The property is documented as READ-ONLY
https://github.com/doctrine/orm/blob/3d9af3187f59930972e7fcb90a1d8059a37b8032/src/Mapping/ClassMetadata.php#L334-L339

VincentLanglet:
The doc need to be updated too, in order to remove reference of BaseUser3 https://github.com/sonata-project/SonataUserBundle/blob/a6169f318ad30a46ced640fad58a686e695e86f0/docs/reference/installation.rst#doctrine-orm-configuration

src/Listener/DoctrineMappingListener.php
VincentLanglet: I think 
```
public function setAttributeOverride(string $fieldName, array $overrideMapping): void
```
should be used instead. (https://github.com/doctrine/orm/blob/3.1.x/src/Mapping/ClassMetadata.php#L1737)

rande: @VincentLanglet https://github.com/doctrine/orm/blob/3.1.x/src/Mapping/ClassMetadata.php#L1765-L1767 it is not possible to change the type using this way. 

This solution is a hack but to be honest having a second class `BaseUser3` just to change the type is not good either. This can end up, with `BaseUser4` etc ...

VincentLanglet: Any recommendation for such situation @greg0ire ? Is it the best way ?

Hanmac:
We probably need to find a way to Migrate the User Data?

My current attempt is this Doctrine Migration file:

```php
$result = $this->connection->executeQuery('SELECT id, roles FROM fos_user_user');
foreach ($result->fetchAllAssociative() as $item) {
    $this->connection->update('fos_user_user', [
        'roles' => json_encode(unserialize($item['roles']))
    ], [
        'id' => $item['id']
    ]);
}

$this->addSql('ALTER TABLE fos_user_user CHANGE roles roles JSON NOT NULL');
```

Hanmac:
@VincentLanglet  @greg0ire how is the status for this MR?

And do we need an update Script for existing data?

VincentLanglet:
> @VincentLanglet @greg0ire how is the status for this MR?
> 
> And do we need an update Script for existing data?

See https://github.com/sonata-project/SonataUserBundle/pull/1685#issuecomment-2113465869

You can do it if you want this to be merged

Hanmac:
@VincentLanglet the Doc update is one thing, should we try to something like my Update Script in a more automated way?

https://github.com/sonata-project/SonataUserBundle/pull/1685#issuecomment-2166019519

Or is the Dev alone with this? (i also don't know if my way is the correct one)

VincentLanglet:
> @VincentLanglet the Doc update is one thing, should we try to something like my Update Script in a more automated way?
> 
> [#1685 (comment)](https://github.com/sonata-project/SonataUserBundle/pull/1685#issuecomment-2166019519)
> 
> Or is the Dev alone with this? (i also don't know if my way is the correct one)

We can document some helper/migrations about how to move from DBAL 3 to DBAL 4 but this cannot be included inside the SonataUserBundle code since we cannot be sure 
- when the user will update his DBAL version
- if he want to execute some data migration or doing something else.

jordisala1991:
I am not sure on what solution to apply here, but right now it is not possible to update to dbal 4, even if userbundle says we are compatible, as soon as you update, the command: doctrine:schema:validate starts complaining:

```
/usr/app $ console doctrine:schema:validate

Mapping
-------

 [FAIL] The entity-class Sonata\UserBundle\Entity\BaseUser mapping is invalid:
 * The field 'Sonata\UserBundle\Entity\BaseUser#roles' uses a non-existent type 'array'.
```

Even if I am not using that mapping file because I switched to the other interface.

PR #1693, opened 2024-08-22
Avoid Username/Email is passed as clear GET Parameter to CheckEmailAction on password reset process

### LINKED ISSUE #1692: Username/Email is passed as clear GET Parameter to CheckEmailAction on password reset process
## Subject

When filling username or email address in Password reset action this value is passed as plain GET parameter to Redirect.

https://github.com/sonata-project/SonataUserBundle/blob/35fe4f6fc49391cc0dff21241acfd49038ee4f5e/src/Action/RequestAction.php#L69-L71

This may cause privacy problems because most server logs are not configured by default to filter or anonymize GET parameters from logfiles.

## Expected results

As there is no need to pass this value as plain value, a minimal invasive solution would be hashing it, before generating the redirect url.

This should NOT be a breaking change, because the only use of this value is the following code:
https://github.com/sonata-project/SonataUserBundle/blob/35fe4f6fc49391cc0dff21241acfd49038ee4f5e/src/Action/CheckEmailAction.php#L37-L39

If you nevertheless think this may be breaking because anyone could have overridden the CheckEmailAction and uses the username value there, i suppose at least to make it configurable if the value is hashed or not.

Please let me know what you think, so i can suppose a pull requst

BA-JBI:
## Subject

Avoid Username/Email is passed as clear GET Parameter to CheckEmailAction on password reset process as discussed in https://github.com/sonata-project/SonataUserBundle/issues/1692

I am targeting this branch, because this fixes an privacy issue and doesn't break anything.

Closes #1692.

### Removed
- After requesting new passwort the username isn't passed to CheckEmailAction anymode

### DIFF
--- a/src/Action/CheckEmailAction.php
+++ b/src/Action/CheckEmailAction.php
@@ -16,16 +16,102 @@ namespace Sonata\UserBundle\Action;
 use Sonata\AdminBundle\Admin\Pool;
 use Sonata\AdminBundle\Templating\TemplateRegistryInterface;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Twig\Environment;
+use TypeError;
 
 final class CheckEmailAction
 {
+
+    private Pool $adminPool;
+    private TemplateRegistryInterface $templateRegistry;
+    private int $tokenTtl;
+
+    /**
+     *  NEXT_MAJOR: Remove `$tokenTtlDeprecated` argument and only allow first types of arguments.
+     *
+     * @param Environment $twig
+     * @param Pool $adminPool
+     * @param TemplateRegistryInterface $templateRegistry
+     * @param int $tokenTtl
+     * @param null $tokenTtlDeprecated
+     */
     public function __construct(
         private Environment $twig,
-        private Pool $adminPool,
-        private TemplateRegistryInterface $templateRegistry,
-        private int $tokenTtl
+        Pool|UrlGeneratorInterface $adminPool,
+        TemplateRegistryInterface|Pool $templateRegistry,
+        int|TemplateRegistryInterface $tokenTtl,
+        ?int $tokenTtlDeprecated = null,
     ) {
+        // NEXT_MAJOR: Remove all checks and use constructor property promotion instead
+        if ($adminPool instanceof UrlGeneratorInterface) {
+            if (!$templateRegistry instanceof Pool) {
+                throw new \TypeError(sprintf(
+                    'Argument 3 passed to %s() must be an instance of %s, %s given.',
+                    __METHOD__,
+                    Pool::class,
+                    \get_class($templateRegistry)
+                ));
+            }
+            $this->adminPool = $templateRegistry;
+
+            if (!$tokenTtl instanceof TemplateRegistryInterface) {
+                throw new \TypeError(sprintf(
+                    'Argument 4 passed to %s() must be an instance of %s, %s given.',
+                    __METHOD__,
+                    TemplateRegistryInterface::class,
+                    \get_class($tokenTtl)
+                ));
+            }
+            $this->templateRegistry = $tokenTtl;
+
+            if (!is_int($tokenTtlDeprecated)) {
+                throw new \TypeError(sprintf(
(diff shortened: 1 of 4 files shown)


Hanmac (approved):

Hanmac (changes requested):
need to update about changed parameters

https://github.com/sonata-project/SonataUserBundle/blob/8713dce8c788b06cd396e6122459768ecad7022f/src/Resources/config/actions_resetting.php#L38-L46

BA-JBI:
Sorry, I forgot the service definitions because I'm already way too spoiled by Symfony's autowiring technology

Hanmac (approved):

Hanmac:
Ugh, I forgot another location again:
https://github.com/sonata-project/SonataUserBundle/blob/8713dce8c788b06cd396e6122459768ecad7022f/src/DependencyInjection/SonataUserExtension.php#L123-L124

Hanmac (approved):

Hanmac:
@VincentLanglet I think this MR is ready to run the Tests?

src/Action/CheckEmailAction.php
```diff
@@ -15,32 +15,21 @@
...
-use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Twig\Environment;
 
 final class CheckEmailAction
 {
     public function __construct(
         private Environment $twig,
-        private UrlGeneratorInterface $urlGenerator,
```
VincentLanglet, line 28: This is a BC break.

If someone does `new CheckEmailAction($twig, $urlGenerator, $adminPool, ...)` it will now crash.

We have to keep a BC signature. You can get inspiration from https://github.com/sonata-project/SonataUserBundle/blob/4.x/src/GoogleAuthenticator/RequestListener.php#L58-L88

```
public function __construct(
    private Environment $twig,
    Pool|UrlGeneratorInterface $adminPool,
    TemplateRegistryInterface|Pool $templateRegistry,
    int|TemplateRegistryInterface $tokenTtl,
    ?int $tokenTtl = null,
) {
    if ($adminPool instanceof UrlGeneratorInterface) {
            if (!$templateRegistry instanceof Pool) {
                throw new TypeError(...);
            }
            $this->adminPool = $templateRegistry;

            @trigger_error(sprintf(
                'Passing an instance of %s as argument 2 to "%s()" is deprecated since sonata-project/user-bundle 5.x and will only accept an instance of %s in version 6.0.',
                UrlGeneratorInterface::class,
                __METHOD__,
                Pool::class
            ), \E_USER_DEPRECATED);
        } else {
            $this->adminPool = $adminPool;
        }
        
        // And so on
}
```

VincentLanglet (changes requested):

Hanmac:
@VincentLanglet about the `native_function_invocation` Lint, should I make an MR for this, or should a bot handle that?

VincentLanglet:
A rebase should solve the issues

BA-JBI:
Tried to apply the backward compatibility as good as possible similar to the suggested code.
Hope this matches the expected solution

src/Action/CheckEmailAction.php
```diff
@@ -15,32 +15,107 @@
...
+                    'Argument 5 passed to %s() must be %s, %s given.',
+                    __METHOD__,
+                    'NULL',
+                    \gettype($tokenTtlDeprecated)
+                ));
+            }
         }
+    }
 
+    public function __invoke(): Response
```
Hanmac, line 117: For BC, the `Request $request` param probably needs to be optional.
(with Deprecate Warning)

VincentLanglet: No I don't think so.

There won't be any error if you pass an extra param

VincentLanglet:
I rebased fixed and merged in https://github.com/sonata-project/SonataUserBundle/pull/1695.

Thanks !

src/Action/CheckEmailAction.php
Hanmac: Ah, i misunderstand the errors there,

it was only Psalm/PHPStan complaining, not a real PHP error.

BA-JBI:
Yey cool
Thank You !!!

PR #1728, opened 2025-11-23
DevKit updates for 5.x branch

SonataCI:


dmaicher:
@VincentLanglet or we merge this and ignore the Symfony 8 CI failure for now :thinking: That is anyway still blocked by https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/1852

Edit: hm but then we would have to update the workflow to not skip/cancel other tests if the Symfony 8 job is failed. Not so easy

VincentLanglet:
> we would have to update the workflow to not skip/cancel other tests

We could change this behavior on dev-kit by using `fail-fast: false` which is maybe smarter indeed.

PR #1739, opened 2026-01-12
User: deprecate eraseCredentials

### LINKED ISSUE #1738: Deprecation: Implementing "UserInterface::eraseCredentials()" is deprecated since Symfony 7.3
## Subject

Class Sonata\UserBundle\Model\User implementing deprecate UserInterface::eraseCredentials()

https://github.com/symfony/symfony/blob/7.4/UPGRADE-7.3.md#security

## Steps to reproduce

Login to admin and use Profiler -> Search profiles -> 302 POST https://.../login_check -> Token -> Logs

## Expected results

No deprecations

## Actual results

> User Deprecated: Since symfony/security-http 7.3: Implementing "App\Entity\SonataUserUser::eraseCredentials()" is deprecated since Symfony 7.3; add the #[\Deprecated] attribute on the method to signal its either empty or that you moved the logic elsewhere, typically to the "__serialize()" method.

```
// src/Entity/SonataUserUser.php
namespace App\Entity;

use Doctrine\DBAL\Types\Types;
use Doctrine\ORM\Mapping as ORM;
use Sonata\UserBundle\Entity\BaseUser3;

#[ORM\Table(name: 'sonata_user__user')]
#[ORM\Entity]
class SonataUserUser extends BaseUser3
{
    #[ORM\Id]
    #[ORM\GeneratedValue]
    #[ORM\Column(name: 'id', type: Types::INTEGER)]
    protected $id;
}
```

Hanmac:
## Subject

I am targeting this branch, because should be BC. (I think)

Closes #1738.

## Changelog

```markdown
### Added

### Changed

### Deprecated
- User::eraseCredentials

### Removed

### Fixed

### Security
```

### DIFF
--- a/src/Document/UserManager.php
+++ b/src/Document/UserManager.php
@@ -50,7 +50,6 @@ final class UserManager extends BaseDocumentManager implements UserManagerInterf
         $password = $this->userPasswordHasher->hashPassword($user, $plainPassword);
 
         $user->setPassword($password);
-        $user->eraseCredentials();
     }
 
     public function findUserByUsername(string $username): ?UserInterface
--- a/src/Entity/UserManager.php
+++ b/src/Entity/UserManager.php
@@ -50,7 +50,6 @@ final class UserManager extends BaseEntityManager implements UserManagerInterfac
         $password = $this->userPasswordHasher->hashPassword($user, $plainPassword);
 
         $user->setPassword($password);
-        $user->eraseCredentials();
     }
 
     public function findUserByUsername(string $username): ?UserInterface
--- a/src/Model/User.php
+++ b/src/Model/User.php
@@ -64,6 +64,7 @@ abstract class User implements LegacyPasswordAuthenticatedUserInterface, UserInt
      */
     public function __serialize(): array
     {
+        $this->plainPassword = null;
         return [
             $this->password,
             $this->salt,
@@ -106,9 +107,9 @@ abstract class User implements LegacyPasswordAuthenticatedUserInterface, UserInt
         }
     }
 
+    #[\Deprecated]
     public function eraseCredentials(): void
     {
-        $this->plainPassword = null;
     }
 
     public function getId()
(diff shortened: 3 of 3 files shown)


Hanmac:
@VincentLanglet is that okay, or are the changes in the UserManager too much?

src/Document/UserManager.php
```diff
@@ -50,7 +50,6 @@ public function updatePassword(UserInterface $user): void
         $password = $this->userPasswordHasher->hashPassword($user, $plainPassword);
 
         $user->setPassword($password);
-        $user->eraseCredentials();
```
dmaicher, line 53: hm but we still support Symfony 6.4 where `eraseCredentials` is not deprecated? :thinking: So I think completely removing the calls might break stuff for some people

src/Model/User.php
```diff
@@ -106,9 +108,9 @@ public function addRole(string $role): void
         }
     }
 
+    #[\Deprecated]
```
dmaicher, line 111: maybe it should be up to the users of this bundle to override the method on their user entity and add this attribute on their end? I feel like we cannot say if people are relying on this method to be called or not.

From what I understand Symfony will stop calling the method in case there is `#[\Deprecated]` present on the method. 

Maybe we should do the same inside the `UserManager` classes? :thinking:

src/Entity/UserManager.php
```diff
@@ -50,7 +50,6 @@ public function updatePassword(UserInterface $user): void
         $password = $this->userPasswordHasher->hashPassword($user, $plainPassword);
 
         $user->setPassword($password);
-        $user->eraseCredentials();
```
dmaicher, line 53: same here

Hanmac:
@dmaicher the problem:

If the method is defined, and not deprecated, then Symfony will complain.
If the method is deprecated, and Sonata UserManager will call them, then PHP 8.4 will complain.

In both cases, it is a dilemma

Hanmac:
@VincentLanglet your opinion to solve this dilemma?

Have our user manager call the method causes the same Deprecation Message I'm trying to avoid

VincentLanglet:
@dmaicher I feel like it could be ok to change this with a note in the Upgrade note/release note that may introduce a minor bc break if someone override the method (?)

dmaicher:
> @dmaicher I feel like it could be ok to change this with a note in the Upgrade note/release note that may introduce a minor bc break if someone override the method (?)

yes I don't have a strong opinion on this. Fine for me.

